PR #34, opened 2022-11-14
download zip package and load field yml from it

aspacca:


pkg/genlib/fields/load.go
```diff
@@ -61,42 +62,99 @@ func makePackageURL(baseURL, integration, version string) (*url.URL, error) {
...
+	r, err = getFromURL(ctx, downloadURL.String())
+	if err != nil {
+		return nil, err
+	}
+
+	h := sha256.New()
+	h.Write([]byte(downloadURL.String()))
+	prefix := hex.EncodeToString(h.Sum(nil))
+
+	packageTempDir, err := os.MkdirTemp(settings.CacheDir(), prefix)
```
scunningham, line 106: Do we need to spool to disk?  Cannot keep zip in RAM?  How big is it?

If continue to spool to disk, perhaps clean up the temp dir on function exit?

pkg/genlib/fields/load.go
```diff
@@ -61,42 +62,99 @@ func makePackageURL(baseURL, integration, version string) (*url.URL, error) {
...
 			return nil, err
 		}
 
-		key := strings.TrimSuffix(filepath.Base(fieldsFileURL), filepath.Ext(fieldsFileURL))
+		fieldsFileContent, err := ioutil.ReadAll(zr)
+		if err != nil {
+			return nil, err
+		}
+
+		_ = zr.Close()
```
scunningham, line 154: Why ignore error on Close()?

pkg/genlib/fields/load.go
```diff
@@ -116,18 +174,25 @@ func getFromURL(ctx context.Context, srcURL string) ([]byte, error) {
 
 	client := &http.Client{}
 	resp, err := client.Do(req)
-
 	if err != nil {
+		defer func(r io.ReadCloser) {
```
scunningham, line 178: This is an odd construct?  defer a function and return immediately?

pkg/genlib/fields/load.go
```diff
@@ -116,18 +174,25 @@ func getFromURL(ctx context.Context, srcURL string) ([]byte, error) {
...
+				_ = r.Close()
+			}
+		}(resp.Body)
 		return nil, err
 	}
 
-	defer resp.Body.Close()
-
 	if resp.StatusCode != http.StatusOK {
+		defer func(r io.ReadCloser) {
```
scunningham, line 187: Same comment as above.

scunningham (commented):
Small nits.

pkg/genlib/fields/load.go
aspacca: > Do we need to spool to disk? Cannot keep zip in RAM? How big is it?

I have to check if zip provide an API for opening a reader

pkg/genlib/fields/load.go
aspacca: once we extracted the fields we don't really care if we cannot close the zip reader, do we? :)

pkg/genlib/fields/load.go
aspacca: yes sorry, the defer is not needed at all, doh! :)

pkg/genlib/fields/load.go
aspacca: did it it ram

scunningham (approved):
Looks good.  Minor style comment.

PR #36, opened 2022-12-05
generator with template

aspacca:


pkg/genlib/config/config.go
```diff
@@ -15,6 +15,7 @@ type ConfigField struct {
 	Fuzziness   int         `config:"fuzziness"`
 	Range       int         `config:"range"`
 	Cardinality int         `config:"cardinality"`
+	Enum        []string    `config:"enum"`
```
aspacca, line 18: new feature, support
```
- name: a.field
  enum: ["pick", "one", "randomly"]
```

pkg/genlib/generator.go
```diff
@@ -250,9 +62,19 @@ func bindConstantKeyword(field Field, fieldMap map[string]emitF) error {
 	return nil
 }
 
-func bindKeyword(field Field, fieldMap map[string]emitF) error {
+func bindKeyword(fieldCfg ConfigField, field Field, fieldMap map[string]emitF) error {
+	if len(fieldCfg.Enum) > 0 {
```
aspacca, line 66: added `enum` support

pkg/genlib/generator_interface.go
```diff
@@ -0,0 +1,289 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
```
aspacca, line 1: here we moved all the common code regardless template usage that was in `generator.go`

pkg/genlib/generator_interface.go
```diff
@@ -0,0 +1,289 @@
...
+
+var (
+	replacer     = strings.NewReplacer(".*", "")
+	keywordRegex = regexp.MustCompile("(\\.|-|_|\\s){1,1}")
+)
+
+// Typedef of the internal emit function
+type emitF func(state *GenState, dupes map[string]struct{}, buf *bytes.Buffer) error
+
+type Generator interface {
```
aspacca, line 58: `Generator` is now an interface: we have to implementation `GenerateJson` (previously `Generate` non-interface) and `GeneratorWithTemplate`

btw, the two concrete implementations probably can be made private

pkg/genlib/generator_with_template.go
```diff
@@ -0,0 +1,529 @@
...
+			templateBuffer.WriteString(fieldTemplate)
+		}
+	}
+
+	return templateBuffer.Bytes()
+}
+
+func NewGeneratorWithTemplate(template []byte, cfg Config, fields Fields) (*GeneratorWithTemplate, error) {
+	if len(template) == 0 {
+		template = generateTemplateFromField(cfg, fields)
```
aspacca, line 69: we can call the factory with an empty template: in this case the template will be generated as json from the `fields`, basically reproducing the same output of the "legacy" generator

this is useful in order to assess regression between legacy and template

when template generator will be performant enough we will remove the legacy and rely on this dynamic template generation

pkg/genlib/generator_with_template.go
```diff
@@ -0,0 +1,529 @@
...
+	}
+
+	tokenizer := regexp.MustCompile(`([^{]*)({{\.[^}]+}})*`)
```
aspacca, line 72: the logic here is to split on `here {{.what.ever0}} there {{.what.ever1}} after`
and have a map of `what.everX` with what must be prefixed (`here `, ` there `) to the random generator value for `what.everX`

this way we can keep writing to the buffer sequentially (prefix and random value), without replace or real template usage: this seems to be the most performant solution so far

replace strategy:
```
Benchmark_Generator
Benchmark_Generator-16                	   13316	     89499 ns/op
Benchmark_GeneratorWithTemplate
Benchmark_GeneratorWithTemplate-16    	      92	  13622401 ns/op
```

write to buffer prefix and random value:
```
Benchmark_Generator
Benchmark_Generator-16                	    9391	    127684 ns/op
Benchmark_GeneratorWithTemplate
Benchmark_GeneratorWithTemplate-16    	    4687	    260068 ns/op
```

we write finally any trailing suffix that has no `.what.everX` following (ie: ` after` in our example)

the logic about `fieldPrefixBuffer` has to be validated by unit tests: this logic should be extracted to a separate function for this purpose

pkg/genlib/generator_with_template.go
```diff
@@ -0,0 +1,529 @@
...
+	fieldMap := make(map[string]emitF)
+	for _, field := range fields {
+		if err := bindField(cfg, field, fieldMap, true); err != nil {
+			return nil, err
+		}
+	}
+
+	// Roll into slice of emit functions
+	emitFuncs := make([]emitF, 0, len(fieldMap))
+	for _, fieldName := range orderedField {
```
aspacca, line 120: `emitFuncs` will be in the order of the fields as they are in the template.
this will allow us to respect the template in `emit()`: without order enforcement the template will be "shuffled"

pkg/genlib/generator_with_template.go
```diff
@@ -0,0 +1,529 @@
...
+		f := fieldMap[fieldName]
+		emitFuncs = append(emitFuncs, f)
+	}
+
+	return &GeneratorWithTemplate{emitFuncs: emitFuncs}, nil
+}
+
+func fieldValueWrapByType(field Field) string {
```
aspacca, line 128: this is utilised only by the dynamic json template generator

pkg/genlib/generator_with_template.go
```diff
@@ -0,0 +1,529 @@
...
+		return fieldValueWrapByType(field)
+	case FieldTypeGeoPoint:
+		return "\""
+	default:
+		return "\""
+	}
+}
+
+func bindConstantKeywordWithTemplate(field Field, fieldMap map[string]emitF) error {
+	prefix := templateFieldMap[field.Name]
```
aspacca, line 158: everywhere, `bind*WithTemplate` difference from `bind*` is
```
prefix := templateFieldMap[field.Name]
```
vs
```
prefix := fmt.Sprintf("\"%s\":", field.Name)
```

we also don't write wrapping quotes, this must be taken care by the template

pkg/genlib/generator_with_template.go
aspacca: final benchmark:
```
Benchmark_Generator-16                     19896             60363 ns/op            3799 B/op         22 allocs/op
Benchmark_GeneratorWithTemplate-16         28335             42998 ns/op             482 B/op         15 allocs/op
```

the difference is from getting rid of [bytes.ReplaceAll](https://github.com/elastic/elastic-integration-corpus-generator-tool/pull/36/commits/87523f82cc5ce235dcebcf93a2463c8ce2cc5718#diff-9e27cdd9458e6fafda86103e129642da8680dd2c3cd87947bb5415f5045bd689L500) in `makeDynamicStubWithTemplate`.

this changes a little the behaviour from legacy generator:
in legacy generator `fields.*` generates from 0 to 5 `fields.keyXYZ`, different for every events. template one generates from 0 to 5 `fields.keyXYZ` as "schema" and they will be the same (with different values) across all events

(indeed the latter is more realistic, at least for the AWS case, where stuff like "dimensions", "labels" and "tags" are usually more the same than less, across a single AWS resource type. things might be different for pods labels maybe, in k8s)

PR #37, opened 2022-12-05
generator with `text/template` package

endorama:
This is a test implementation leveraging `text/template` as template engine for data generation.


pkg/genlib/generator_with_template.go
```diff
@@ -0,0 +1,593 @@
...
+	// trailingTemplate = fieldPrefixBuffer
+	// // Preprocess the fields, generating appropriate emit functions
+	// fieldMap := make(map[string]emitF)
+	// for _, field := range fields {
+	// 	if err := bindField(cfg, field, fieldMap, objectKeys, templateFieldMap); err != nil {
+	// 		return nil, err
+	// 	}
+	// }
+
+	// for k := range objectKeysFields {
```
aspacca, line 219: this is required for when you have `field.*` or similar: [here](https://github.com/elastic/elastic-integration-corpus-generator-tool/pull/37/files#diff-9e27cdd9458e6fafda86103e129642da8680dd2c3cd87947bb5415f5045bd689R47-R85) we transform that to 0 from 5 `field.keyN`, directly in the the `templateFieldMap`. since you'll get rid of it you have to find a way to mimic the same logic somewhere else.

beware that this indeed is only true if we want to generate a default json template given a fields definition. that would be required only if we get rid of the legacy generator.

endorama:
My latest benchmark is:
```
❯ benchstat mer\ \ 7\ dic\ 2022,\ 10.22.46,\ CET.txt
name                          time/op
_Gen2-12                      64.8µs ± 9%
_Gen3-12                      15.3µs ± 7%
_Generator-12                 39.9µs ± 6%
_GeneratorWithTemplate-12     39.3µs ± 8%
_GeneratorWithTemplateAWS-12  8.74µs ± 7%

name                          alloc/op
_Gen2-12                      10.8kB ± 0%
_Gen3-12                      2.94kB ± 0%
_Generator-12                   482B ± 0%
_GeneratorWithTemplate-12       482B ± 0%
_GeneratorWithTemplateAWS-12    160B ± 0%

name                          allocs/op
_Gen2-12                         219 ± 0%
_Gen3-12                        97.0 ± 0%
_Generator-12                   15.0 ± 0%
_GeneratorWithTemplate-12       15.2 ± 5%
_GeneratorWithTemplateAWS-12    5.00 ± 0%
```

`Gen2` does not have acceptable performances (CPU is okish but memory is not).  
`Gen3` is performant enough (better than `Generator` CPU wise but costs more memory),  considering that offers the flexibility of using `text/template` but with the drawback of not always empty `GenState` or `dupes`.

We can evaluate other template engine, there are "compile to go" ones that have extreme performance improvements. The trade off here would be the need to implement templates before hand, but for known cases it's worth a try.

ruflin:
I did just run this and I was slightly surprised by the destionation:

```
./elastic-integration-corpus-generator-tool generate-with-template assets/templates/aws.vpcflow/vpcflow.log assets/templates/aws.vpcflow/vpcflow.fields.yml --tot-size=2000
File generated: /Users/ruflin/Library/Application Support/elastic-integration-corpus-generator-tool/corpora/1670945614-vpcflow.log
```
The binary itself is inside `ruflin/Dev/...`. What does it push to `Application Support` directory?

assets/templates/aws.vpcflow/vpcflow.conf.yml
```diff
@@ -0,0 +1,25 @@
+- name: Version
```
ruflin, line 1: This file does not have to be specified when running the tool. Is it picked up by convention? Why is it separate from the fields.yml? In case different configs should be specified?

pkg/genlib/generator_with_template.go
endorama: I added support for this, resolving the conversation.

assets/templates/aws.vpcflow/vpcflow.conf.yml
endorama: This file has to be manually passed to the generator via the `--config-file` flag https://github.com/elastic/elastic-integration-corpus-generator-tool/blob/main/cmd/generate.go#L91

Support for this is introduced in #36 (on top of which this PR is based).

endorama:
@ruflin I'm not sure about the folder, this PR should not change that behaviour. I got surprised too when I run it, but I didn't dig into why it behaves like that.

aspacca:
> The binary itself is inside `ruflin/Dev/...`. What does it push to `Application Support` directory?

I borrowed xdg usage from the changelog tool from @endorama (that I used to scaffold this project)  :)

the default path the the corpa location is here https://github.com/elastic/elastic-integration-corpus-generator-tool/blob/main/internal/settings/settings.go#L32-L34

it can be customised by env variable as like here: https://github.com/elastic/elastic-integration-corpus-generator-tool/blob/main/internal/settings/xdg_test.go#L99

we can change the behaviour in case

pkg/genlib/gen2.go
```diff
@@ -0,0 +1,101 @@
...
+			// FIXME: this writes directly to buf, while template expects a return value
+			_ = fieldMap[field](state, dupes, buf)
+			// NOTE: mandatory empty string return to avoid cluttering the template, otherwise
+			// you get a "<no value>" for each field written to buf
+			return ""
+		}
+
+		// NOTE: Funcs needs to be run before parsing. As we want to bind state and dups this needs
+		// to happen here, parsing the template for each emit.
+		// This is not needed and if we can get rid of state (which I've not seen used in value
```
aspacca, line 72: `dups` can be removed, it was used in the legacy `makeDynamicStub`, but it's no longer used anywhere

`state.prevCache` is used for cardinality, fuzziness and keyword
`state.pool` as buffer cache in the `makeDynamicStub`

pkg/genlib/gen3.go
```diff
@@ -0,0 +1,85 @@
...
+
+package genlib
+
+import (
+	"bytes"
+	"strings"
+	"text/template"
+)
+
+// Gen2 is resolved at construction to a slice of emit functions
```
aspacca, line 13: `Gen3` :)

pkg/genlib/gen3.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+// NewGen2
+//
+// From current benchmarks this generator is at least as CPU performant as Generator, while providing templating.
+// Is worse in terms of RAM used and memory allocations (but working on template parse optimization may help).
+//
+// If you're getting a nil pointer dereference like error, something like:
+// template: generator:2:2: executing "generator" at <generate "foobar">: error calling generate: runtime error: invalid memory address or nil pointer dereference
+// is probably due to the "foobar" not being a valid field. In this case the generate function will try accessing fieldMap at an invalid location.
```
aspacca, line 25: can we solve this with `if fieldFromMap, ok := fieldMap[field]; ok`?

pkg/genlib/gen3.go
```diff
@@ -0,0 +1,85 @@
...
+		return nil, err
+	}
+
+	e := Gen3{}
+	e.tpl = parsedTpl
+
+	return &e, nil
+}
+
+func (e Gen3) Emit(state *GenState, buf *bytes.Buffer) error {
```
aspacca, line 78: we could remove `state` at all here

but indeed with this implementation we have to be sure that `state` at https://github.com/elastic/elastic-integration-corpus-generator-tool/pull/37/files#diff-4615405939260828cc99730784d91e4a0a671b44d80b92095390f0094653720aR64 get the proper... state (sic!) :)

pkg/genlib/testdata/template.tpl
```diff
@@ -0,0 +1 @@
+{{generate "aws.vpcflow.version"}} {{generate "aws.vpcflow.account_id"}} {{generate "aws.vpcflow.interface_id"}} {{generate "aws.vpcflow.pkt_srcaddr"}} {{generate "aws.vpcflow.pkt_dstaddr"}} {{generate "source.port"}} {{generate "destination.port"}} {{generate "network.transport"}} {{generate "network.packets"}} {{generate "network.bytes"}} {{generate "event.start"}} {{generate "event.end"}} {{generate "aws.vpcflow.action"}} {{generate "aws.vpcflow.log_status"}}
```
aspacca, line 1: can we add the logic for packets and log status here?
in order to compare the benchmark we should generate the same data as in https://github.com/elastic/elastic-integration-corpus-generator-tool/blob/047ef7f2f5b59b7a918c283f26e8622309e5c0e4/pkg/genlib/generator_test.go#L805

pkg/genlib/generator_test.go
```diff
@@ -54,81 +229,111 @@ func Test_Cardinality(t *testing.T) {
 }
 
 func test_CardinalityT[T any](t *testing.T, ty string) {
-
+	template := `{"alpha":"{{.alpha}}"}`
```
aspacca, line 232: all these tests do not to seem to be updated to cover `Gen2` and `Gen3` (unless I missed something), we should do :)

PR #38, opened 2022-12-06
Support javascript expression

aspacca:


assets/templates/aws.vpcflow/vpcflow.conf.yml
```diff
@@ -0,0 +1,40 @@
...
+- name: Packets
+  range: 1048576
+- name: Bytes
+  expression: >
+    function process() {
+      return {{.Packets}} * 15;
+    }
+- name: Action
+  enum: ["ACCEPT", "REJECT"]
+- name: LogStatus
```
ruflin, line 27: Very interesting idea. So this bascially the value of this fields depends on values of other fields. Does the used field have to show up in the final event? 

For later: Would be nice to have a playground / sample linked for easy testing of the expressions for devs.

aspacca: > So this bascially the value of this fields depends on values of other fields. Does the used field have to show up in the final event?

yes, they have to be present in the template, so that a value for them is generated, and you can reference only other fields present in the template before the current one that depends on them

this is a big limitation, and while I prefer having the more advanced feature given from the javascript stdlib, without the needs to add functions to the template if we use go template, paired with the worst performace of the javascript engine, I would go with the go template.

in the latter I can do something like the following: 
```
{{ $packets := packets }}{{ if eq $packets 0 }} NODATA {{ else }} {{logStatus}} {{ end }}
```

Iwe can just assign all the variables to the generated values in the beginning of the template and we have a workaround for the the limitation about referring only "previous rendered" field (I could assign a variable and never print it)

Benchmark using the js engine:

```
_Generator-16                 3.28µs ± 0%
_GeneratorUsingExpression-16   367µs ± 0%

name                          alloc/op
_Generator-16                 2.43kB ± 0%
_GeneratorUsingExpression-16   320kB ± 0%

name                          allocs/op
_Generator-16                   21.0 ± 0%
_GeneratorUsingExpression-16   3.96k ± 0%

```

PR #40, opened 2022-12-21
Templates benchmark

aspacca:
### Data schemas

When collecting data with Elastic Agent and shipping to Elasticsearch, there are 4 different data schemas. This is important as the data schemas look different and we must align during generation on what data schemas we are talking about. In the diagram below, the schemas A, B, C and D are shown:

```mermaid
flowchart LR
    D[Data/Endpoint]
    EA[Elastic Agent]
    ES[Elasticearch]
    IP[Ingest Pipeline]
    D -->|Schema A| EA -->|Schema B| IP -->|Schema C| ES -->|Schema D| Query;
```

* **Schema A**: This is the schema the Elastic Agent collects. It could be a line in a log file, response of an http request, syslog event etc. The Elastic Agent input knows how to handle this structure.
* **Schema B**: This is the schema the Elastic Agent ships to Elasticsearch (to the ingest pipeline). This is a JSON document which contains all the processing the Elastic Agent did on schema A. For example the content of the log line is the in the `message` field and meta information around the host was added to the event.
* **Schema C**: In case an ingest pipeline exists, the ingest pipeline converts schema B to schema C. This can be taking apart a log message with grok or enriching data with geoip. If there is no ingest pipeline, schema B and C are equal. 
* **Schema D**: This is the schema users write queries on in Elasticsearch. Schema C can be different from D in the scenario of runtime fields, otherwise C and D are equal.

Schema C is the one that is  defined in integration packages in the `fields.yml` files for each data stream. In the following, our **foucs** is on **Schema B** with mentions of schema A.

### Introduction:
Currently the corpus generator can produce only `Schema C`
We want to support generating `Schema A` and `Schema B`

We figured out that in order to do so we need a way to generate data based on a template.
A few investigating effort were made about this:
- #36 a first attempt to introduce a custom template, similar to golang text template syntax, supporting only replacing fields placeholder.
- #37 a second attempt to introduce support for golang text template engin: this will allow for some scripting features that will pair the features available in spigot (see: https://github.com/leehinman/spigot/blob/main/pkg/generator/aws/vpcflow/vpcflow.go#L99-L107)
- #38 a parallel attempt to move the script features to the config definition: a field can define a js function that is evaluated at runtime. a field can access the value of the fields already generated before as they are order in the template

### Analysis:
[...]

PR #41, opened 2023-01-03
With template

aspacca:
Done:
-  [x] added `generate-with-template` command
-  [x]  support for `--template-type`: `placeholder` (custom gotext-like template supporting only placeholders) and `gotext` (standard lib go text_template package)

Todo:
-  [x] add basic math operation in `gotext`

Notes:
- the branch keeps the whole history of the various branches where general refactoring and the benchmarking of different template engine were done. those branches' PRs were closed and we keep here partial history of the process
- finale choice was on go text template rather than jet as for https://github.com/CloudyKit/jet/issues/203


aspacca:
@scunningham you can check benchmark results at https://github.com/elastic/elastic-integration-corpus-generator-tool/pull/40

cmd/generate_with_template.go
```diff
@@ -0,0 +1,82 @@
...
+			}
+
+			fmt.Println("File generated:", payloadFilename)
+
+			return nil
+		},
+	}
+
+	generateWithTemplateCmd.Flags().StringVarP(&configFile, "config-file", "c", "", "path to config file for generator settings")
+	generateWithTemplateCmd.Flags().StringVarP(&templateType, "template-type", "y", "placeholder", "either `placeholder` only or full `gotext` template")
```
endorama, line 79: I propose to change this description a bit to clarify the two possible options. I think we also need more documentation around these values.

```suggestion
	generateWithTemplateCmd.Flags().StringVarP(&templateType, "template-type", "y", "placeholder", "either 'placeholder' or 'gotext'")
```

cmd/generate_with_template.go
```diff
@@ -0,0 +1,82 @@
...
+	generateWithTemplateCmd.Flags().StringVarP(&totSize, "tot-size", "t", "", "total size of the corpus to generate")
```
endorama, line 80: Can we specify the "unit" for this total size? Is a line count or bytes, MB, GB? Is there a way to specify the unit?

internal/corpus/generator.go
```diff
@@ -20,6 +21,13 @@ import (
 	"github.com/spf13/afero"
 )
 
+const (
+	templateTypeCustom = iota
+	templateTypeGoText
+)
+
+var ErrNotValidTemplate = errors.New("please, pass --template-type as one of `placeholder` or `gotext`")
```
endorama, line 29: Replacing backticks with `'` to avoid escaping and rendering issues in terminals.
```suggestion
var ErrNotValidTemplate = errors.New("please, pass --template-type as one of 'placeholder' or 'gotext'")
```

endorama (commented):
A couple of minor comments but looks good to me!

endorama:
@aspacca I saw there is no documentation around how to use `placeholder` and `gotext` templating engines. Should we add it with this PR? Like how to use them (how to write a template and/or config file, what features are available, etc)

cmd/generate_with_template.go
aspacca: yes! we use https://pkg.go.dev/github.com/dustin/go-humanize#ParseBytes on the value of `tot-size`

aspacca:
> I saw there is no documentation around how to use `placeholder` and `gotext` templating engines. Should we add it with this PR? Like how to use them (how to write a template and/or config file, what features are available, etc)

indeed: please check if the content I've added is enough

aspacca:
@endorama ready to merge?

endorama (approved):

PR #47, opened 2023-01-25
Aws.sqs schema.b

aspacca:
Create the artifacts required to generate schema B data for `aws.sqs` dataset in the `aws` package


assets/templates/aws.sqs/sqs.gotext.log
```diff
@@ -0,0 +1,2 @@
...
...": {{ generate "EventDuration" }}, "agent_id_status": "verified", {{ $eventIngested := generate "EventIngested" }} "ingested": "{{ $eventIngested.Format "2006-01-02T15:04:05.999999Z07:00" }}", "module": "aws", "dataset": "aws.sqs" }, "aws": { "cloudwatch": { "namespace": "AWS/SQS" }, "dimensions": { "QueueName": "{{ generate "QueueName" }}" }, "sqs": { "metrics": { "ApproximateAgeOfOldestMessage": { "avg": {{ generate "OldestMessageAge" }} }, "ApproximateNumberOfMessagesDelayed": { "avg": {{ generate "Delayed" }} }, "ApproximateNumberOfMessagesNotVisible": { "avg": {{ generate "NotVisible" }} }, "ApproximateNumberOfMessagesVisible": { "avg": {{ generate "Visible" }} }, "NumberOfMessagesDeleted": { "avg": {{ generate "Deleted" }} }, "NumberOfMessagesReceived": { "avg": {{ generate "Received" }} }, "NumberOfMessagesSent": { "avg": {{ generate "Sent" }} }, "NumberOfEmptyReceives": { "avg": {{ generate "EmptyReceives" }} }, "SentMessageSize": { "avg": {{ generate "SentMessageSize" }} }, } }, "tags": { "createdBy": {{ generate "TagsCreatedBy" }}, } }, "cloud": { "account": { "id": "000000000000", "name": "elastic-observability" }, "provider": "aws", "region": {{ generate "Region" }} } }
```
ruflin, line 2: I wonder if there is a way that we could allow users to build is template in "pretty json" to make it much more readable, simplify PR reviews etc. and only on generation munch it into a signle line?

assets/templates/aws.sqs/sqs.gotext.log
```diff
@@ -0,0 +1,2 @@
+{ "create" : { "_index": "metrics-aws.sqs-default" } }
```
ruflin, line 1: Does everyone that builds a template have to add this? I would assume we can automatically add this based on the knowledge from the dataset or if not, have a config that links to the different templates for generation and has it in.

This might also solve the issue below. If there is just 1 event per doc and we know it is json (could be indicated) we can munch it into a single line afterwards.

assets/templates/aws.sqs/sqs.gotext.log
aspacca: I see the value of building a "pretty json" template, but we have to consider that a template can be something else than a json, and having multiple lines can be semantically meaningful in a different context.

We have to provide a flag somewhere (when running the command/in some config file/etc) if a specific template must be "folded" in a single line or not, if we want to provide such option

ruflin: yes. My thinking is that "somewhere" there will be config for all the generation task and the templates linked to it. One way to indicate it could also be the file ending.

assets/templates/aws.sqs/sqs.gotext.log
aspacca: > Does everyone that builds a template have to add this?

Definitely no: I just added it here in the draft (while it might stick for a few iterations)
Two considerations:
1. I still have to investigate if that line is required when loading a bulk request payload from a locally file in rally
2. Once we will have support for different outputs, knowing that we will generate rally output will add the line automatically

assets/templates/aws.sqs/sqs.placeholder.log
```diff
@@ -0,0 +1,2 @@
...
...": 300000, "name": "cloudwatch" }, "event": { "duration": {{ .EventDuration }}, "agent_id_status": "verified", "ingested": "{{ .EventIngested }}", "module": "aws", "dataset": "aws.sqs" }, 	"aws": { 		"cloudwatch": { 			"namespace": "AWS/SQS" 		}, 		"dimensions": { 			"QueueName": "{{ .QueueName }}" 		}, 		"sqs": { 			"metrics": { 				"ApproximateAgeOfOldestMessage": { 					"avg": {{ .OldestMessageAge }} 				}, 				"ApproximateNumberOfMessagesDelayed": { 					"avg": {{ .Delayed }} 				}, 				"ApproximateNumberOfMessagesNotVisible": { 					"avg": {{ .NotVisible }} 				}, 				"ApproximateNumberOfMessagesVisible": { 					"avg": {{ .Visible }} 				}, 				"NumberOfMessagesDeleted": { 					"avg": {{ .Deleted }} 				}, 				"NumberOfMessagesReceived": { 					"avg": {{ .Received }} 				}, 				"NumberOfMessagesSent": { 					"avg": {{ .Sent }} 				}, 				"NumberOfEmptyReceives": { 					"avg": {{ .EmptyReceives }} 				}, 				"SentMessageSize": { 					"avg": {{ .SentMessageSize }} 				}, 			} 		}, 		"tags": { 			"createdBy": {{ .TagsCreatedBy }}, 		} 	}, 	"cloud": { 		"account": { 			"id": "000000000000", 			"name": "elastic-observability" 		}, 		"provider": "aws", 		"region": {{ .Region }} 	} }
```
endorama, line 2: What do you think about making `account.name` and `account.id` generated fields but with a static value?

aspacca: those value are going to be repeated a lot across multiple template.
I think that if we use hardcoded value it will be easier to implement some kind of common template snippets to include in multiple templates.
once we'll work on that feature we can revise, but for the moment I'd keep so

aspacca:
@endorama . this is ready for final review, please let me know if we can merge, thanks!

endorama (approved):

PR #51, opened 2023-02-01
generate in goroutines

aspacca:
`main` branch:
```
goos: darwin
goarch: amd64
pkg: github.com/elastic/elastic-integration-corpus-generator-tool/pkg/genlib
cpu: Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
Benchmark_GeneratorCustomTemplateJSONContent-16    	  235588	     47286 ns/op	     432 B/op	      14 allocs/op
Benchmark_GeneratorTextTemplateJSONContent-16      	   24507	    487438 ns/op	   48725 B/op	    2252 allocs/op
Benchmark_GeneratorCustomTemplateVPCFlowLogs-16    	 6558933	      1840 ns/op	      64 B/op	       2 allocs/op
Benchmark_GeneratorTextTemplateVPCFlowLogs-16      	  526669	     22920 ns/op	    2322 B/op	      95 allocs/op
PASS
ok  	github.com/elastic/elastic-integration-corpus-generator-tool/pkg/genlib	57.004s
```

this branch:
```
goos: darwin
goarch: amd64
pkg: github.com/elastic/elastic-integration-corpus-generator-tool/pkg/genlib
cpu: Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
Benchmark_GeneratorCustomTemplateJSONContent-16    	   97422	    145236 ns/op	   18980 B/op	     972 allocs/op
Benchmark_GeneratorTextTemplateJSONContent-16      	   18703	    820674 ns/op	   49159 B/op	    2265 allocs/op
Benchmark_GeneratorCustomTemplateVPCFlowLogs-16    	 1504225	      8486 ns/op	     501 B/op	      35 allocs/op
Benchmark_GeneratorTextTemplateVPCFlowLogs-16      	  299754	     43765 ns/op	    2540 B/op	     111 allocs/op
PASS
ok  	github.com/elastic/elastic-integration-corpus-generator-tool/pkg/genlib	73.641s
```

PR #58, opened 2023-02-14
Highlight field errors in generation when they happen

endorama:
This PR aims at helping template developers in 2 situations:
1. when using `generate` on a missing field
2. when an error occurs when generating data for a field

In case `generate()` is being used in the template on a missing field
the current behaviour of returning an empty string hides the problem to
the template developer.

As it is not expected for the template to try generating a missing
field, this commit returns a known string highlighting the problem.

In case there in an error in the bind function for generating a value
for a field, instead of returning an empty string, return a fixed known
string that highlight the issue.

### DIFF
--- a/pkg/genlib/generator_with_text_template.go
+++ b/pkg/genlib/generator_with_text_template.go
@@ -40,7 +40,7 @@ func NewGeneratorWithTextTemplate(tpl []byte, cfg Config, fields Fields) (*Gener
 	templateFns["generate"] = func(field string) interface{} {
 		bindF, ok := fieldMap[field]
 		if !ok {
-			return ""
+			return "<missing field>"
 		}
 
 		value, err := bindF(state, nil)
--- a/pkg/genlib/generator_with_text_template.go
+++ b/pkg/genlib/generator_with_text_template.go
@@ -45,7 +45,7 @@ func NewGeneratorWithTextTemplate(tpl []byte, cfg Config, fields Fields) (*Gener
 
 		value, err := bindF(state, nil)
 		if err != nil {
-			return ""
+			return "<unvalued field>"
 		}
 		return value
 	}
--- a/pkg/genlib/generator_with_text_template.go
+++ b/pkg/genlib/generator_with_text_template.go
@@ -45,7 +45,7 @@ func NewGeneratorWithTextTemplate(tpl []byte, cfg Config, fields Fields) (*Gener
 
 		value, err := bindF(state, nil)
 		if err != nil {
-			return "<unvalued field>"
+			return ""
 		}
 		return value
 	}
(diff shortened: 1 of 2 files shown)


pkg/genlib/generator_with_text_template.go
```diff
@@ -40,12 +40,12 @@ func NewGeneratorWithTextTemplate(tpl []byte, cfg Config, fields Fields) (*Gener
 	templateFns["generate"] = func(field string) interface{} {
 		bindF, ok := fieldMap[field]
 		if !ok {
-			return ""
+			return "<missing field>"
```
aspacca, line 43: I was thinking about this, I'd rather send a `null`, so if the value is assigned to a variable that's expected to be of another type rather than a string, and further functions are applied the template will most likely fail to render

the best would be if we could avoid any kind of failure in the `generate` body

`value, err := bindF(state, nil)` can be indeed refactored without returning any error (see [here](https://github.com/elastic/elastic-integration-corpus-generator-tool/pull/51/files#diff-0fc87380e8715fdb1a33c600e6513325ba82801e849851c5856064b7beac59b8R94)

we cannot do anything about `bindF, ok := fieldMap[field]`, but maybe let's fail the whole command, since having in a template `generate "fieldNotInFieldsYaml` is not a transient error but a bug in the template

endorama: > I'd rather send a null

Returning `nil` does not halt execution, just renders an empty template.

> but maybe let's fail the whole command,

I tried panicking from `generate` function, but `template.Execute` recover from panics (see [here](https://github.com/golang/go/blob/0fa9ba8abeb4519db6fd5121e2794d8818daf8dc/src/text/template/exec.go#L204-L220), go1.19).  

I've not been able to have the panic triggered in `generate` to escape this recovery, due to this I've found no way to properly halt execution.

> value, err := bindF(state, nil)

I will remove the commit from this PR, we will address this in a future update.

aspacca: > I tried panicking from `generate` function, but `template.Execute` recover from panics (see [here](https://github.com/golang/go/blob/0fa9ba8abeb4519db6fd5121e2794d8818daf8dc/src/text/template/exec.go#L204-L220), go1.19).
> 
> I've not been able to have the panic triggered in `generate` to escape this recovery, due to this I've found no way to properly halt execution.

we can have an error channel and close it on `!ok`, then in the emit() we add a select with `<-errorChan` after parsing the template, that returns an error, and a default that `return nil`

aspacca: > we can have an error channel and close it on `!ok`, then in the `emit()` we add a select with `<-errorChan` after parsing the template, that returns an error, and a default that `return nil`

please, @endorama , check my proposal: https://github.com/elastic/elastic-integration-corpus-generator-tool/pull/51/commits/2e1bc5e5f0e58273288a6ed3928fe40f37809792#diff-0fc87380e8715fdb1a33c600e6513325ba82801e849851c5856064b7beac59b8R120

endorama:
Closing this as the requested improvements are part of #51

PR #59, opened 2023-02-14
Enable all sprig functions in templates

endorama:
Non hermetic functions are defined as "impure" functions. The list is
available here: https://github.com/Masterminds/sprig/blob/581758eb7d96ae4d113649668fa96acc74d46e7f/functions.go#L70

We enable them mainly due to date related functions that are very
beneficial to template developers.

### DIFF
--- a/pkg/genlib/generator_with_text_template.go
+++ b/pkg/genlib/generator_with_text_template.go
@@ -31,7 +31,7 @@ func NewGeneratorWithTextTemplate(tpl []byte, cfg Config, fields Fields) (*Gener
 
 	state := NewGenState()
 
-	templateFns := sprig.HermeticTxtFuncMap()
+	templateFns := sprig.TxtFuncMap()
 
 	templateFns["timeDuration"] = func(duration int64) time.Duration {
 		return time.Duration(duration)


aspacca (approved):

pkg/genlib/generator_with_text_template.go
```diff
@@ -31,7 +31,7 @@ func NewGeneratorWithTextTemplate(tpl []byte, cfg Config, fields Fields) (*Gener
 
 	state := NewGenState()
 
-	templateFns := sprig.HermeticTxtFuncMap()
+	templateFns := sprig.TxtFuncMap()
 
 	templateFns["timeDuration"] = func(duration int64) time.Duration {
```
aspacca, line 36: I'd would be nice if we can add `timeDuration` directly to sprig in the future

aspacca:
I think you need to merge main on the branch @endorama

pkg/genlib/generator_with_text_template.go
endorama: I can work in submitting this function to the library, but is not clear to me which is the expected usage. `sprig` already provides `dateModify` function, how would this be different?

aspacca: > dateModify, mustDateModify
> The dateModify takes a modification and a date and returns the timestamp.

it is unclear to me if it returns a `time.Time`, what's exactly a "timestamp"? :)
we can test anyway, if so, no need, since we can have 
```
{{- $durationInSecond := generate "durationInSecond" }}
{{- $durationString := cat $durationInSecond "s" | nospace }}
{{- $modifiedTime := $timeField | dateModify $durationString }}
{{ $modifiedTime.Format "2006-01-02T15:04:05.999999Z07:00" }}
```

otherwise we need a function to generate a `time.Duration` from an int or string (I could not find any in sprig), since we don't have anything in the fields type that generates a `time.Duration` (and we probably cannot have, since the type comes from package spec)

PR #65, opened 2023-03-01
Add `aws.ec2_logs` schema-b template

endorama:
Add schema-b templates for `aws.ec2_logs`

### DIFF
--- a/assets/templates/aws.ec2_logs/schema-b/fields.yml
+++ b/assets/templates/aws.ec2_logs/schema-b/fields.yml
@@ -1,0 +1,8 @@
+- name: process.name
+  type: keyword
+- name: event.ingested
+  type: date
+- name: aws.ec2.ip_address
+  type: ip
+- name: message
+  type: text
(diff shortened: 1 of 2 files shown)


assets/templates/aws.ec2_logs/schema-b/configs.yml
```diff
@@ -0,0 +1,18 @@
+- name: process.name
+  enum: ["journal", "kernel", "systemd"]
+- name: agent.id
+  value: "da6cb4c8-c84c-4c5f-97c7-f8586a098af4"
+- name: cloud.region
+  enum: ["us-east-1", "us-east-2", "us-west-1", "us-west-2", "ap-south-1", "ap-northeast-3", "ap-northeast-2", "ap-southeast-1", "ap-southeast-2", "ap-northeast-1", "ca-central-1", "eu-central-1", "eu-west-1", "eu-west-2", "eu-west-3", "eu-north-1", "sa-east-1", "af-south-1", "ap-east-1", "ap-south-2", "ap-southeast-3", "eu-south-2", "eu-central-2", "me-south-1", "me-central-1"]
```
aspacca, line 6: nit: if we don't give the same cardinality to `cloud.region`, the same `host.name` could end up in different regions

aspacca: yes, 1/100 is correct

aspacca (approved):

PR #66, opened 2023-03-06
Clean code

aspacca:
This PR change the interface of the generator, that now expects the total size of events to generate.
We decide the number of events to generate dividing the total size request by the size of a single initial event that's discarded.
Requesting a 0 total size means that the number of events to generate is infinite.

In the PR there's also a general clean-up of the code

### DIFF
--- a/pkg/genlib/generator_with_custom_template.go
+++ b/pkg/genlib/generator_with_custom_template.go
@@ -6,14 +6,23 @@ package genlib
 
 import (
 	"bytes"
+	"io"
 	"regexp"
 )
 
-var trailingTemplate []byte
+type emitter struct {
+	fieldName string
+	fieldType string
+	emitFunc  emitFNotReturn
+	prefix    []byte
+}
 
 // GeneratorWithCustomTemplate is resolved at construction to a slice of emit functions
 type GeneratorWithCustomTemplate struct {
-	emitFuncs []emitFNotReturn
+	totEvents        uint64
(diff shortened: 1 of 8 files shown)


pkg/genlib/generator_with_custom_template.go
```diff
@@ -110,12 +167,18 @@ func (gen GeneratorWithCustomTemplate) Emit(state *GenState, buf *bytes.Buffer)
 }
 
 func (gen GeneratorWithCustomTemplate) emit(state *GenState, buf *bytes.Buffer) error {
-	for _, f := range gen.emitFuncs {
-		if err := f(state, buf); err != nil {
-			return err
+	if gen.totEvents == 0 || state.counter < gen.totEvents {
```
endorama, line 170: @aspacca is `state.counter` updated somewhere? I've not been able to find it.

aspacca: @endorama yes, in the `Emit()`

https://github.com/elastic/elastic-integration-corpus-generator-tool/blob/c9537381950cc004c4249e30debefa819427dfba/pkg/genlib/generator_with_custom_template.go#L164
and https://github.com/elastic/elastic-integration-corpus-generator-tool/blob/c9537381950cc004c4249e30debefa819427dfba/pkg/genlib/generator_with_text_template.go#L145

it's not in the diff because nothing changed there :)

endorama (approved):

PR #80, opened 2023-03-16
Fuzziness as float and range constraint

### LINKED ISSUE #76: Refactor `fuzziness` and let it comply to `range` constraint
The `fuzziness` setting in the config was intended to be able to generate values for numeric fields that vary by a given delta from the previous generated value.
Initially the value was defined as maximum variation in percentage from the previous generated value: the variation will be randomly applied as increasing or decreasing. This means that if you have a `fuzziness` of "10" and the first value generated is 50, the second value will be generated with the constraint `45 < second value < 55`.

In https://github.com/elastic/elastic-integration-corpus-generator-tool/pull/46 we refactored `fuzziness` in order to be able to set a custom `denominator`, so that we were not forced to maximum variation interval as a percentage: this was coherent with the same refactoring we did in the same PR for `cardinality`, but it's not really functional to anything but for the fact to express a percentage that's not a integer.

Having to express both a `numerator` and a `denominator` just for expressing not integer percentage can be more easily solved expressing `fuzziness` as a float. Once you express the `fuzziness` as float, you can use values lower than "1", opening to the possibility to have a delta in permille or any other lower magnitude order than 1% (ie: `fuzziness: 0.01`)

My proposal is to refactor the `fuzziness` field once again in order to accept a single float value.

`fuzziness` setting will look like the following:
```
fuzziness: 0.1
```

A second topic is letting `fuzziness` comply to `range` constraint.

Currently the value generated applying `fuzziness` setting does not respect the `min` and `max` constraint defined in `range` for the same field.
Imagine the following config:
```
fuzziness: 50
range:
  min: 30.
  max: 40.
```

The first value will be generated with the following constraint: `30. < X < 40.`
Assuming that the generated X for the first value will be "35.", the second value will be generated with the following constraint: `17.5 < X < 52.5`, that can potentially result in values "escaping" the `range` constraint.

There are different ways we can implement `fuzziness` compliance to `range`, I will make a brief proposal of the easiest one I can think about.

Currently we calculate and "adjust ratio" starting from fuzziness: `AR = rand(0.0,1.0) * fuzziness`
Randomly we decide if the delta from the previous value will be negative or positive and we either add 1. to AR or remove AR to 1.:
In pseudo code:
```
AR = rand(0.0,1.0) * FUZZINESS # AR = 0.7
DIRECTION = rand_choice([NEGATIVE, POSITIVE])
if DIRECTION == NEGATIVE:
    AR = 1. - AR # AR = 0.3
ELSE:
    AR += 1. # AR = 1.7

NEW_VALUE = PREVIOUS_VALUE * AR
```

The new logic will be:
```
LOWER_AR_BOUND = PREVIOUS_VALUE * (1 - FUZZINESS)
IF LOWER_AR_BOUND < RANGE_MIN:
    LOWER_AR_BOUND = RANGE_MIN
HIGHER_AR_BOUND = PREVIOUS_VALUE * (1 + FUZZINESS)
IF HIGHER_AR_BOUND > RANGE_MAX:
    HIGHER_AR_BOUND = RANGE_MAX
NEW_VALUE = rand(LOWER_AR_BOUND , HIGHER_AR_BOUND)
```

cc @endorama, @ruflin

aspacca:
`fuzziness` field is now a single float value that expressed as `1 == 100%` variation delta from the previous value
values generated by applying `fuzziness` now respects defined `range` constraint, if present

closes #76

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,16 +208,12 @@ It is possible to tweak the randomness of the generated data through a config fi
 ##### Sample config
 ```yaml
 - name: aws.dynamodb.metrics.AccountMaxReads.max
-  fuzziness:
-    numerator: 1
-    denominator: 10
+  fuzziness: 0.1
   range:
     min: 0
     max: 100
 - name: aws.dynamodb.metrics.AccountMaxTableLevelReads.max
-  fuzziness:
-    numerator: 1
-    denominator: 20
+  fuzziness: 0.05
   range:
     min: 0
     max: 50
@@ -225,9 +221,7 @@ It is possible to tweak the randomness of the generated data through a config fi
     numerator: 1
     denominator: 20
 - name: aws.dynamodb.metrics.AccountProvisionedReadCapacityUtilization.avg
-  fuzziness:
-    numerator: 1
-    denominator: 10
+  fuzziness: 0.1
 - name: aws.cloudwatch.namespace
   cardinality:
     numerator: 1
@@ -254,7 +248,7 @@ It is possible to tweak the randomness of the generated data through a config fi
 The config file is a yaml file consisting of an array of config entry.
 For each config entry the following fields are available
 - `name` *mandatory*: dotted path field
-- `fuzziness` *optional (`long` and `double` type only)*: delta from the previous generated value for the same field, expressed as a ratio between a `numerator` and a `denominator`
+- `fuzziness` *optional (`long` and `double` type only)*: maximum delta from the previous generated value for the same field, expressed as float where 1 = 100%
 - `range` *optional (`long` and `double` type only)*: value will be generated between `min` and `max`
 - `cardinality` *optional*: distribution of different values for the field, expressed as a ratio between a `numerator` and a `denominator`
 - `object_keys` *optional (`object` type only)*: list of field names to generate in a object field type. if not specified a random number of field names will be generated in the object filed type.
--- a/README.md
+++ b/README.md
@@ -248,7 +248,7 @@ It is possible to tweak the randomness of the generated data through a config fi
 The config file is a yaml file consisting of an array of config entry.
 For each config entry the following fields are available
 - `name` *mandatory*: dotted path field
-- `fuzziness` *optional (`long` and `double` type only)*: maximum delta from the previous generated value for the same field, expressed as float where 1 = 100%
+- `fuzziness` *optional (`long` and `double` type only)*: when generating data you could want generated values to change in a known interval. Fuzziness allow to specify the maximum delta a generated value can have from the previous value (for the same field), as a delta percentage; value must be between 0.0 and 1.0, where 0 is 0% and 1 is 100%. When not specified there is no constraint on the generated values, boundaries will be defined by the underlying field type
 - `range` *optional (`long` and `double` type only)*: value will be generated between `min` and `max`
 - `cardinality` *optional*: distribution of different values for the field, expressed as a ratio between a `numerator` and a `denominator`
 - `object_keys` *optional (`object` type only)*: list of field names to generate in a object field type. if not specified a random number of field names will be generated in the object filed type.
--- a/pkg/genlib/config/config.go
+++ b/pkg/genlib/config/config.go
@@ -18,6 +18,7 @@ type Ratio struct {
 }
 
 type Range struct {
+  // NOTE: we want to distinguish when Min/Max are explicitly set to zero value or are not set at all. We use a pointer, such that when not set will be `nil`.
 	Min *float64 `config:"min"`
 	Max *float64 `config:"max"`
 }
--- a/pkg/genlib/config/config.go
+++ b/pkg/genlib/config/config.go
@@ -1,10 +1,16 @@
 package config
 
 import (
-	"github.com/elastic/go-ucfg/yaml"
+	"errors"
+
 	"io/ioutil"
+	"math"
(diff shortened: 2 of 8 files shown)


README.md
```diff
@@ -254,7 +248,7 @@ It is possible to tweak the randomness of the generated data through a config fi
 The config file is a yaml file consisting of an array of config entry.
 For each config entry the following fields are available
 - `name` *mandatory*: dotted path field
-- `fuzziness` *optional (`long` and `double` type only)*: delta from the previous generated value for the same field, expressed as a ratio between a `numerator` and a `denominator`
+- `fuzziness` *optional (`long` and `double` type only)*: maximum delta from the previous generated value for the same field, expressed as float where 1 = 100%
```
aspacca, line 251: > maximum delta from the previous generated value for the same field, expressed as float where 1 = 100%

please, @endorama , help me to find a better explanation for this, especially the `1 = 100%` part :)

pkg/genlib/config/config.go
```diff
@@ -22,14 +28,46 @@ type Config struct {
...
+	Fuzziness   float64  `config:"fuzziness"`
 	Range       Range    `config:"range"`
 	Cardinality Ratio    `config:"cardinality"`
 	Enum        []string `config:"enum"`
 	ObjectKeys  []string `config:"object_keys"`
 	Value       any      `config:"value"`
 }
 
+func (r Range) MinAsInt64() (int64, error) {
+	if r.Min == nil {
```
aspacca, line 40: we want to differ when `Min/Max` are explicitly set to zero value or are not set at all.
The only was is to make the value a pointer, so if not set the zero value of the pointer will be `nil` and we can differentiate from explicitly being set to the zero value of the underlying pointer type

pkg/genlib/config/config.go
```diff
@@ -1,19 +1,25 @@
...
+
 type Ratio struct {
 	Numerator   int `config:"numerator"`
 	Denominator int `config:"denominator"`
 }
 
 type Range struct {
-	Min interface{} `config:"min"`
-	Max interface{} `config:"max"`
+	Min *float64 `config:"min"`
```
aspacca, line 21: let's parse as `float64` and convert to `int64` with getters (`MinAsInt64`/`MaxAsInt64`) instead of doing [magic](https://github.com/elastic/elastic-integration-corpus-generator-tool/pull/80/files#diff-eeffd89b6f2d2b3db8542f09093ea87c32181bfa90f4dfba8141118acfa0320dL226-L233) when reading the value to guess the original type.

pkg/genlib/config/config.go
```diff
@@ -1,19 +1,25 @@
...
+	"errors"
+
 	"io/ioutil"
+	"math"
 	"os"
+
+	"github.com/elastic/go-ucfg/yaml"
 )
 
+var rangeBoundNotSet = errors.New("range bound not set")
```
aspacca, line 13: error returned by the getters to indicate that the value was not set at all

pkg/genlib/generator_interface.go
```diff
@@ -559,21 +531,22 @@ func bindLong(fieldCfg ConfigField, field Field, fieldMap map[string]any) error
 		return nil
 	}
 
+	min, _ := fieldCfg.Range.MinAsFloat64()
```
aspacca, line 534: we don't need to check for the value not being set here, since `min` will be returned as 0, and we'll take the maximum between `0` and `lowerBound>=0` and `max` will be `math.MaxFloat64` and we'll take the minimum between  `math.MaxFloat64` and `higherBound<=math.MaxFloat64`

pkg/genlib/config/config.go
endorama, line 21:
```suggestion
  // NOTE: we want to distinguish when Min/Max are explicitly set to zero value or are not set at all. We use a pointer, such that when not set will be `nil`.
	Min *float64 `config:"min"`
```

pkg/genlib/generator_interface.go
```diff
@@ -559,21 +531,22 @@ func bindLong(fieldCfg ConfigField, field Field, fieldMap map[string]any) error
...
 		dummyInt := dummyFunc()
 		if previousDummyInt, ok := state.prevCache[field.Name].(int); ok {
-			adjustedRatio := rand.Float64() * (fuzzinessNumerator / fuzzinessDenominator)
-			if rand.Int()%2 == 0 {
-				adjustedRatio += 1.
-			} else {
-				adjustedRatio = 1. - adjustedRatio
-			}
-			dummyInt = int(math.Ceil(float64(previousDummyInt) * adjustedRatio))
+			lowerBound := float64(previousDummyInt) * (1 - fieldCfg.Fuzziness)
```
endorama, line 541: I see this code repeated with small differences in multiple places. Can we extract it to a single function and performs only conversion where needed?

README.md
endorama:
```suggestion
- `fuzziness` *optional (`long` and `double` type only)*: when generating data you could want generated values to change in a known interval. Fuzziness allow to specify the maximum delta a generated value can have from the previous value (for the same field), as a delta percentage; value must be between 0.0 and 1.0, where 0 is 0% and 1 is 100%. When not specified there is no constraint on the generated values, boundaries will be defined by the underlying field type
```

endorama (approved):

PR #82, opened 2023-03-23
Move and improve docs

endorama:
With this PR i completed the initial work of improving and structuring the tool documentation.

I moved most of the content out of the README in separate files, and added some ancillary content.


docs/field-configurations.md
```diff
@@ -9,13 +9,53 @@ They must be added to a file named `configs.yml` in the assets template folder o
...
 - `range` *optional (`long` and `double` type only)*: value will be generated between `min` and `max`
 - `cardinality` *optional*: distribution of different values for the field, expressed as a ratio between a `numerator` and a `denominator`
 - `object_keys` *optional (`object` type only)*: list of field names to generate in a object field type; if not specified a random number of field names will be generated in the object filed type
 - `value` *optional*: hardcoded value to set for the field (any `cardinality` will be ignored)
-- `enum` *optional (`keyword` type only)*: list of strings to randomly chose from a value to set for the field
+- `enum` *optional (`keyword` type only)*: list of strings to randomly chose from a value to set for the field (any `cardinality` will be ignored)
```
endorama, line 18: @aspacca I'm not sure this addition is correct.

aspacca: @endorama indeed not
we generate the values exclusively from the `enum` list, and we apply `cardinality`

what can happens is that the length of the `enum` list is shorter than the `cardinality` values
for example:
```
enum: ["1", "2", "3"]
cardinality:
   numerator: 1
   denominator: 6
```

according to `cardinality` we'll generate a value 6 times, each value will be 1 out of 3 available values in the `enum` list

so the generated values will be repeated and not 6 different ones.
but still once we generate the 6 values (with duplicates) we will emit the 6 values always in the same sequence.

The latest is a details about `cardinality` that's probably worth to clarify (I still have to read your new content, so maybe it's already covered): `cardinality` not only tries to produce a `numerator/denominator` amount of different values[1] but it does in a sequence that will be repeated in a fixed way while emitting the events
so if `numerator/denominator` is 3 and the cardinal sequence generated is "B is a random value", "random value 0", "a different random value", and we emit 7 events, the resulting events will be
1. event: "B is a random value"
2. event: "random value 0"
3. event: "a different random value"
4. event: "B is a random value"
5. event: "random value 0"
6. event: "a different random value"
7. event: "B is a random value"

[1] as much of its possibility, that's not always possible like the case above or if we don't have enough "entropy" to generate all the number of values requested different from each other - imagine asking to generate 20 different integer in the range between 1 and 10, or generating more different keyword type values that the number of different values that we can ask github.com/Pallinder/go-randomdata to generate

CONTRIBUTING.md
```diff
@@ -0,0 +1,28 @@
...
+
+`$ make build`
+
+# Running tests
+
+`$ make test`
+
+# Adding License header
+
+All files in this repository should have the License header. To do this, use:
```
aspacca, line 17: nit: all files that not reside in the `assets` folder

docs/field-configurations.md
```diff
@@ -9,13 +9,53 @@ They must be added to a file named `configs.yml` in the assets template folder o
 The config file is a yaml file consisting of an array of config entry.
 
 For each config entry the following fields are available:
-- `name` *mandatory*: dotted path field, as in `fields.yml`
```
aspacca, line 12: why removing "as in `fields.yml`", it's relevant that the fields defined in the config file must match an entry in the fields.yml, otherwise they will be ignored :)

docs/usage.md
```diff
@@ -0,0 +1,36 @@
...
+**Example**:
+
+```shell
+$ go run main.go generate aws dynamodb 1.14.0 -t 1000 --config-file config.yml
+File generated: /path/to/corpora/1649330390-aws-dynamodb-1.14.0.ndjson
+```
+
+# Generate schema-b data from a template
+
+To do this, use the `generate-with-template` command. This command targets a specific template within the `assets/templates` folder.
```
aspacca, line 22: > This command targets a specific template within the `assets/templates` folder.

```
A body of templates, fields definition and fields definition configuration are already available in the `assets/templates` folder. You can rely on them or write your own ones.
Please consider opening a PR adding your custom templates, fields definition and fields definition configuration to the existing body, if they belong to an integration, so that we can enrich the catalogue for everyone.
```

docs/glossary.md
```diff
@@ -0,0 +1,35 @@
...
+
+# Fields definition
+
+Datasets define data structure. Within them there are fields definitions that describe the fields a dataset provides. 
+
+In the context of this tool we can refer to:
+- the fields definition within a dataset in an integration package
+- a file named `fields.yml` that contains fields definition that is not (yet) part of a package
+
+# Fields definition configuration
```
aspacca, line 23: I have to say: having `Fields definition` and `Fields definition configuration` is a little confusing for me :)

I'd prefer some more specific and different definition, to create less confusion, what about
`Fields definition` and `Fields configuration`?

docs/usage.md
aspacca: I'd say it does not really targets what's in the `assets/templates` folder, like the same way the `generate` command targets a specific dataset within an integration package at a specific version :)

See my suggestion above

aspacca:
@endorama please let me know if you have time polishing this PR addressing my comments or you prefer I take over it, thanks :)

aspacca (approved):

aspacca:
@v1v any idea why the required actions are waiting? thanks

v1v:
> @v1v any idea why the required actions are waiting? thanks

I guess those are the former GitHub checks from Jenkins

image


Hence they shoiuld be removed, and likely added the new GitHub action checks. Let me do it

IN addition I'd say to rebase this PR with `main`

aspacca:
> I guess those are the former GitHub checks from Jenkins

uhm, no, I've just added them. also tried with `test` (the job) as github action

maybe I forgot to merge `main` indeed

v1v:
I see now... it's filtering by docs! Let me put the workaround

endorama:
Hey @aspacca I missed the notification of your review on this PR. Overall :+1: on the changes you did.

PR #88, opened 2023-04-12
Adding initial template for kubernetes Pod

gizas:
I have created the initial template for kubernetes pod datastream. This tempate produces data ready to be ingested from rally tracks.
See below the generated file comparing to real data extracted from cluster:

[GENERATED-gotext.json.txt](https://github.com/elastic/elastic-integration-corpus-generator-tool/files/11212277/GENERATED-gotext.json.txt)

[Initial_from_realPOD.json.txt](https://github.com/elastic/elastic-integration-corpus-generator-tool/files/11212278/Initial_from_realPOD.json.txt)

**Command to run:**
```yaml
/elastic-integration-corpus-generator-tool generate-with-template ./assets/templates/kubernetes.pod/gotext.tpl ./assets/templates/kubernetes.pod/fields.yml -c ./assets/templates/kubernetes.pod/configs.yml -y gotext -t 1000
```

### Features to implement:
- Produce dynamically the kubernetes.pod fileds from enumaration choices 👍 Done
- Create relevant container template 👍 Done
- Test with rally and compare results with existing real corpus 👍 Done. See doc [here](https://github.com/elastic/elastic-integration-corpus-generator-tool/pull/88#issuecomment-1536082638)

### Findings after testing:

1.  **Timestamps** of  generated data are spread within 1 hour from the time of the triggering the tool. Although generated data can be used for indexing, for the needs of visualisations return empty responses if queries are for time window larger than 1h. Next steps is to try to create the timestamps by using range (or cardinality) functions and spread them to multiple hours

2. The generator tool produces output in multiple lines. The rally tool needs **each entry in one line**. For now this issue is minor as we provide our templates as one liners: https://github.com/elastic/elastic-integration-corpus-generator-tool/pull/88/files#diff-44eae17c43b58d9d956a9c89b53eed2aa72ef46ad7e10400f26a0a61cd22ccfcR17. 

  - Also elastic-package tool dump command that will use the generator tool will dump the data in one line! Needs to be tested

3. For every Rally run we need to generate the corpus data (eg. https://github.com/elastic/rally-tracks/pull/373/files#diff-dbafff74aad306950d4c38f30c7612f06cae89395c58311d54ca26a2c374fc03R52) and the mappings of the indices we test (eg. https://github.com/elastic/rally-tracks/pull/373/files#diff-0b2bc88dee0704c8bae38dbe5719417945216348c62273f09940d1afcb7a7eea). The need of corpus generation is matched here, but we still dont have a fully **automated way to generate the mapping templates** based on the relevant package version we test every time. We make the assumption that mappings dont change so often and can be extracted from any given cluster but still manual process is needed. Issue is not a blocker


gizas:
Pod and container templates have been created:

For container datastream:
```yaml
./elastic-integration-corpus-generator-tool generate-with-template ./assets/templates/kubernetes.container/gotext.tpl ./assets/templates/kubernetes.container/fields.yml -c ./assets/templates/kubernetes.container/configs.yml -y gotext -t 1000
```

For pod:
```yaml
 ./elastic-integration-corpus-generator-tool generate-with-template ./assets/templates/kubernetes.pod/gotext.tpl ./assets/templates/kubernetes.pod/fields.yml -c ./assets/templates/kubernetes.pod/configs.yml -y gotext -t 1000
```

aspacca:
@gizas all good here?

gizas:
> @gizas all good here?

Yes my tests are ok and I have managed to produce the wanted outcome. I will open this pr once I fully test it with rally track.
I need to find time to run a full test and will come back to you

gizas:
@aspacca , @martijnvg FYI:
https://github.com/elastic/observability-dev/blob/generatortool/docs/infraobs/cloudnative-monitoring/dev-docs/elastic-generator-tool-with-rally.md

I have created the needed corpus and 55G files takes less than couple of minutes !!!! Lets see it in action

aspacca:
@gizas we merged in master a change in the format of the fields generation config file
from
```
- name: cloud.availabilit_zone
  value: "europe-west1-d"  
- name: agent.id
  value: "12f376ef-5186-4e8b-a175-70f1140a8f30"
```

to

```
fields:
  - name: cloud.availabilit_zone
    value: "europe-west1-d"  
  - name: agent.id
    value: "12f376ef-5186-4e8b-a175-70f1140a8f30"
```

we'll later introduce something like 
```
formatter:
  - strip_newlines
fields:
```


so that you won't need to duplicate anymore the templates, like you do in this PR, and just have a single one "pretty-printed" that will be emitted with newlines stripped

please, feel free to suggest how the `formatter` "concept" should look like, thanks

aspacca:
> https://github.com/elastic/observability-dev/blob/generatortool/docs/infraobs/cloudnative-monitoring/dev-docs/elastic-generator-tool-with-rally.md

👍 

please, let's align on what's currently available in `elastic-package`: https://github.com/elastic/elastic-package/blob/main/docs/howto/generate_corpus.md#generate-a-rally-track-for-a-package-dataset-and-run-a-rally-benchmark

assets/templates/kubernetes.container/configs.yml
```diff
@@ -0,0 +1,49 @@
...
+- name: event.duration
+  range:
+    min: 1
+    max: 4000000
+- name: faults
+  range:
+    min: 0
+    max: 500000
+- name: kubernetes.container.rootfs.inodes.used
+  cardinality:
```
aspacca, line 24: is it correct to have `cardinality` here instead of `range`?

btw, `cardinality` is going to be refactored at https://github.com/elastic/elastic-integration-corpus-generator-tool/pull/92

you'll have to change your configuration to `cardinality: 100000`

assets/templates/kubernetes.container/gotext.tpl
```diff
@@ -0,0 +1,17 @@
...
...B2-07-B9-51-C7", "82-BA-19-5A-58-82", "82-D7-1D-53-29-32", "82-DE-CE-61-02-10", "82-ED-72-F5-A1-33", "86-00-ED-A3-E1-AD", "8A-79-72-4A-60-4A", "8A-E3-4A-82-34-25", "92-A0-37-79-0E-53", "96-BE-33-2C-70-9D", "96-CF-87-1F-7A-92", "96-FA-6D-ED-B3-AB", "9E-95-61-2E-65-35", "9E-C1-CC-21-8C-D2", "A2-37-A4-B9-EB-C7", "A2-53-05-15-D5-C8", "A2-7D-AA-82-43-98", "A6-22-4E-BB-59-76", "A6-3A-53-19-D4-FA", "A6-CC-39-88-CE-C4", "AA-9B-77-AE-BC-C6", "AA-9C-F4-F0-07-64", "AE-11-3C-D5-0D-25", "AE-54-A8-A4-81-1F", "B2-33-C7-E9-AB-FA", "B6-BB-E9-22-E9-C4", "B6-C3-EB-BE-0A-01", "BA-6F-65-B6-5F-2E", "BA-87-AE-57-BF-E7", "BA-8A-04-F8-D8-4E", "BA-AE-1E-A0-3E-10", "BE-07-E2-A4-AE-01", "C2-5B-7B-C3-6B-CB", "C2-8C-52-6A-DF-C4", "C2-94-4E-05-70-5A", "C6-94-B2-D7-B3-4E", "CA-98-EB-50-A5-59", "CE-91-F0-0E-1E-4B", "CE-97-6C-58-40-E2", "D2-DB-58-48-AF-07", "D6-5F-99-9F-D6-D8", "D6-C7-6A-AB-24-95", "DA-06-9A-33-79-D7", "DA-DA-DC-F5-83-84", "DA-E4-2F-50-6E-E0", "E2-35-F1-C8-81-A0", "E2-50-4E-47-58-E2", "E2-6B-7C-F4-90-35", "E6-DE-78-5C-0A-AA", "EE-EC-71-F6-7F-16", "F2-45-C9-18-35-EA", "F2-9E-A0-1D-FA-66", "F6-06-54-94-B3-A3", "F6-AF-C9-EC-1A-8A", "FE-6A-D4-63-97-8D", "FE-F1-90-50-28-AD" ], "architecture":"x86_64"}}}
```
aspacca, line 17:
```
"name":"{{ generate "container.name"}}", ..., "name":"{{ generate "container.name" }}"
```

beware that this will result in generating two different values from the `container.name` `enum`. so you will end up with a different `container.name` and `kubernetes.container.name`, not sure if that's what you want.

in general, every value that must be the same, every time is "printed" in a single event should generated only once

assets/templates/kubernetes.container/gotext_multiline.tpl
```diff
@@ -0,0 +1,554 @@
...
+         "usage": {{divf $pct 1000000}}
+      },
+      "id":"{{ $container_uId }}"
+   },
+   "kubernetes": {
+      "container":{
+         "start_time":"{{$timestamp.Format "2006-01-02T15:04:05.999999Z07:00"}}",
+         "memory":{
+            "rss":{
+               "bytes": {{generate "Bytes"}}
```
aspacca, line 33: `"bytes": {{generate "Bytes"}}`

not sure if it's a concern for you, but I'd guess that the sum of the various `memory` stats will not be "correct" (meaning more usage than available or similar)

assets/templates/kubernetes.container/gotext_multiline.tpl
```diff
@@ -0,0 +1,554 @@
...
+                  "pct": {{divf $pct 1000000}}
+               }
+            },
+            "pagefaults": "{{ $faults }}"
+         },
+         "rootfs":{
+            "inodes":{
+               "used": {{ generate "kubernetes.container.rootfs.inodes.used" }}
+            },
+            "available":{
```
aspacca, line 60: `{{generate "Bytes"}}`

not sure if it's a concern for you, but I'd guess that the sum of the various `rootfs` stats will not be "correct" (meaning used+available not matching capacity or similar)

assets/templates/kubernetes.container/gotext_multiline.tpl
gizas: For now is not concern, we can generate dummy values

assets/templates/kubernetes.container/gotext_multiline.tpl
gizas: Yes no concern at the moment

assets/templates/kubernetes.container/configs.yml
gizas: Agree changed to range

assets/templates/kubernetes.container/gotext.tpl
gizas: Changed as well. You are right better like this although not affecting our tests

gizas:
> Next steps is to try to create the timestamps by using range (or cardinality) functions and spread them to multiple hours

I have addressed this issue by manually produsing timestamps

Also relevant PR introduces the period flag that can be used in the future
https://github.com/elastic/elastic-integration-corpus-generator-tool/commit/62a8465ba70aab20191b25dca6e6d797a7ab60fe


> The generator tool produces output in multiple lines. The rally tool needs each entry in one line. 
We have created onliner versions of templates in this PR

Also the [elastic-package-benchmark-generate-corpus command](https://github.com/elastic/elastic-package#elastic-package-benchmark-generate-corpus) will output results in one line per doc entry

> Generation of Rally Templates (as part of generatol tool)

Is not needed anymore, as I have tested the ` elastic-package dump installed-objects --package kubernetes` command that can extract index templates. Relevant instructons added inside the [Readme of TSDB2 rally track](https://github.com/elastic/rally-tracks/blob/7c3f611d1b8df6194f90dbb31bd6ebbda0076713/tsdb2/README.md#generation-of-templates)

aspacca:
> 1. **Timestamps** of  generated data are spread within 1 hour from the time of the triggering the tool. Although generated data can be used for indexing, for the needs of visualisations return empty responses if queries are for time window larger than 1h. Next steps is to try to create the timestamps by using range (or cardinality) functions and spread them to multiple hours

this should be addressed by #95 :)

> 2. The generator tool produces output in multiple lines. The rally tool needs **each entry in one line**. For now this issue is minor as we provide our templates as one liners: https://github.com/elastic/elastic-integration-corpus-generator-tool/pull/88/files#diff-44eae17c43b58d9d956a9c89b53eed2aa72ef46ad7e10400f26a0a61cd22ccfcR17.
> * Also elastic-package tool dump command that will use the generator tool will dump the data in one line! Needs to be tested

yes, elastic-package does it. still elastick-package use v0.5.0 of the corpus generator, so the assets in this PR won't be compatible for the moment. see next point comment for further details

> 3. For every Rally run we need to generate the corpus data (eg. https://github.com/elastic/rally-tracks/pull/373/files#diff-dbafff74aad306950d4c38f30c7612f06cae89395c58311d54ca26a2c374fc03R52) and the mappings of the indices we test (eg. https://github.com/elastic/rally-tracks/pull/373/files#diff-0b2bc88dee0704c8bae38dbe5719417945216348c62273f09940d1afcb7a7eea). The need of corpus generation is matched here, but we still dont have a fully **automated way to generate the mapping templates** based on the relevant package version we test every time. We make the assumption that mappings dont change so often and can be extracted from any given cluster but still manual process is needed. Issue is not a blocker

similarly to what's done in this [elastic-package PR](https://github.com/elastic/elastic-package/pull/1232), we'll add a `benchmark rally` command that will install the relevant assets of the local package. this will come with releasing v0.6.0 of the corpus generator and upgrading the dependency in elastic-package

assets/templates/kubernetes.container/configs.yml
```diff
@@ -0,0 +1,56 @@
...
+  - name: Percentage
+    range:
+      min: 0
+      max: 5000
+    fuzziness: 0.005
+  - name: container.name
+    enum: ["web", "default-http-backend", "dnsmasq", "csi-driver", "web", "web", "web", "prometheus", "konnectivity-agent", "sidecar", "kubedns", "metrics-server-nanny", "web", "web", "fluentbit", "autoscaler", "gke-metrics-agent", "elastic-agent", "web", "kube-state-metrics", "metrics-server", "fluentbit", "elastic-agent", "web", "prometheus-to-sd-exporter"]
+  - name: timedate
+    # Add more dates below in case you want your data to be spread
+    enum: ["2023-05-15" , "2023-05-16"]
```
aspacca, line 53: you can replace `timedate` and `timehour` with a single:
```
- name: timestamp
  period: 24h
```

assets/templates/kubernetes.container/fields.yml
```diff
@@ -0,0 +1,42 @@
...
+  type: keyword   
+- name: agent.id
+  type: keyword
+- name: agent.name
+  type: keyword
+- name: agent.version
+  type: keyword
+- name: agent.ephemeral_id
+  type: keyword  
+- name: container.network.ingress.bytes
```
aspacca, line 15: any reason for having this as `double` and then casting to `int` in the template?
same for `container.network.egress.bytes`, `Offset` and `faults`

assets/templates/kubernetes.container/gotext_multiline.tpl
```diff
@@ -0,0 +1,559 @@
...
+{{- $container_uId := uuidv4 }}
+{{- $suffix := split "-" $uId }}
+{{- $timestamp := generate "timestamp" }}
+{{- $fulltimestamp := $timestamp.Format "2006-01-02T15:04:05.999999Z07:00" }}
+{{- $resttime := split ":" $fulltimestamp }}
+{{- $picktimedate := generate "timedate" }}
+{{- $timehour := generate "timehour" }}
+{{- $offset := generate "Offset" | int }}
+{{- $faults := generate "faults" | int }}
+{{- $pct := generate "Percentage" | float64 }}
```
aspacca, line 19: if you `divf $pct` you should not need to cast it to `float64`: https://github.com/Masterminds/sprig/blob/master/numeric.go#L178

assets/templates/kubernetes.container/configs.yml
gizas: Yes I thought of that but at the time of latest tests the PR was not yet merged.

Do you think is a problem to let them like this now?

assets/templates/kubernetes.container/gotext_multiline.tpl
gizas: Fixed in 31b2839e62064fc4d20c24fd78079e161efa8d06

assets/templates/kubernetes.container/fields.yml
gizas: Fixed in 31b2839e62064fc4d20c24fd78079e161efa8d06

assets/templates/kubernetes.container/configs.yml
gizas: Also I have my dates fixed as the rally track will run some visualisations that have queries in specific timerange (see here https://github.com/elastic/rally-tracks/pull/373/files#diff-18fad8cb6e0990bf9839cad4594f45f8cfde9d2a0a8cce9495614f74c5ab670cR2)

aspacca: if the rally tracks have fixed dates I'd leave it as it is.
it's not a problem per se, indeed it seems it is intended because of the rally tracks, but it's just my personal preference of having as less "hardcoded" stuff as possible :)

aspacca (approved):

assets/templates/kubernetes.pod/gotext.tpl
```diff
@@ -0,0 +1,18 @@
...
+{{- $agentId := generate "agent.id" -}}
+{{- $agentVersion := generate "agent.version" -}}
+{{- $agentName := generate "agent.name" -}}
+{{- $agentEphemeralid := generate "agent.ephemeral_id" -}}
+{{- $rxbytes := generate "container.network.ingress.bytes" -}}
+{{- $txbytes := generate "container.network.egress.bytes" -}}
+{{- $uId := uuidv4 -}}
+{{- $pod_uId := uuidv4 -}}
+{{- $suffix := split "-" $uId -}}
+{{- $offset := generate "Offset" | int -}}
```
aspacca, line 16: nit: `Offset` still casted to `int`

assets/templates/kubernetes.container/gotext.tpl
```diff
@@ -0,0 +1,19 @@
...
+{{- $timehour := generate "timehour" -}}
+{{- $agentId := generate "agent.id" -}}
+{{- $agentVersion := generate "agent.version" -}}
+{{- $agentName := generate "agent.name" -}}
+{{- $agentEphemeralid := generate "agent.ephemeral_id" -}}
+{{- $uId := uuidv4 -}}
+{{- $pod_uId := uuidv4 -}}
+{{- $container_uId := uuidv4 -}}
+{{- $suffix := split "-" $uId -}}
+{{- $offset := generate "Offset" | int -}}
```
aspacca, line 15: nit: `Offset` and `faults` belowe, still casted to `int`

gizas: fixed in [1f0eec3](https://github.com/elastic/elastic-integration-corpus-generator-tool/pull/88/commits/1f0eec3b7cb21974560340daa4edae971a77465d)

assets/templates/kubernetes.pod/gotext.tpl
gizas: fixed in [1f0eec3](https://github.com/elastic/elastic-integration-corpus-generator-tool/pull/88/commits/1f0eec3b7cb21974560340daa4edae971a77465d)

PR #92, opened 2023-05-10
Scalar cardinality value

### LINKED ISSUE #75: Refactor `cardinality`
The `cardinality` setting in the config was intended to be able to generate repeated values for a given field.
Initially the value was defined as a permille of the total number of events generated: this means that if you have a `cardinality` of "10" and "1000" events generated, only "10" different values will be generated for a given field, and if you generated "2000" events, "20" different values will be generated for the same field.
With the refactoring happened in https://github.com/elastic/elastic-integration-corpus-generator-tool/pull/15 the behaviour changed: the value was indicating the total different values to generate as a ratio between the `cardinality` value and "1000": meaning that "10" will generate "100" different values, and "20" will generated "50" different values, regardless the number of total events generated.

In https://github.com/elastic/elastic-integration-corpus-generator-tool/pull/46 we refactored `cardinality` in order to be able to set a custom `denominator`, so that we were not forced to express a ratio based on "1000": this was functional to the fact that you might need to express relationship between different fields cardinalities that might not have a zero reminder when divided by "10" (it's specifically meaningful when you have cardinality on fields with an enum different from 2 and 5)

Still this refactoring did not introduced back the option to express the cardinality dynamically as a factor of the total number of events generated.

My proposal is to refactor the `cardinality` field once again in order to express both the original behaviour and the new one (both are useful in my opinion and we should provide all of them).

`cardinality` setting will now accept two different schemas:
1.
```
cardinality:
  fixed: X
```

This is just a different way of expressing the current behaviour: "generate X different values for the field". There's no need to express this value as a ratio between a numerator and a denominator.
If the cardinality is expressed this way we will generate X amount of different values for the field. If the number of events generated is less than X, we will generate less different values.
The X values are generated sequentially: every X events the values sequence will repeat itself. If the number of total events generated divided by X has no zero reminder, the last repeated sequence will contain less than X values.

2.
```
cardinality:
  numerator: X
  denominator: Y
```

This is the current schema, but it will change the behaviour: "generate X/Y of the total number of events different values".
If the cardinality is expressed this way we will generate N amount of different values for the field where `N = (tot events / Y) * X`.
If you set "10" as `numerator` and "100" as `denominator` and you generate "1000" total events, "100" different values will be generated. With the same settings, generating "1500" total events, will produce "150" different values.
If the number of events to generate is infinite the behaviour will be the following: "generate X/Y different values for the field", basically reproducing the actual behaviour and falling back to the behaviour of 1., just expressing X as X/Y.
If the formula `N = (tot events / Y) * X` returns anything `<1` we generate only 1 different value.
The N values are generated sequentially: every N events the values sequence will repeat itself. If the number of total events generated divided by N has no zero reminder, the last repeated sequence will contain less than N values. This might happen since N might not be an integer and we will floor N to the nearest integer.

Note:
> If the number of total events generated divided by N has no zero reminder, the last repeated sequence will contain less than N values. This might happen since N might not be an integer and we will floor N to the nearest integer.

We might change the way we calculate the total number of events to generate so that we make them with zero reminder if divided by each of the Ns (as 2.) and Xs (as 1.) resulting as parsing the config. This means that you cannot generate less events than the minimum required to fulfil every cardinality configured.
We might decide on this later on: the refactoring proposed in this issue can proceed without defining this detail. In case we will adopt nothing should change in the code dealing with cardinality, but just in the functions (one for each template engine) that calculate the total number of events to be generated.

cc @endorama, @ruflin

aspacca:
fixes #75 

cardinality now is an integer value defining how many different values to generate for a given field

### DIFF
--- a/docs/fields-configuration.md
+++ b/docs/fields-configuration.md
@@ -12,7 +12,7 @@ For each config entry the following fields are available:
 - `name` *mandatory*: dotted path field, matching an entry in [Fields definition](./glossary.md#fields-definition)
 - `fuzziness` *optional (`long` and `double` type only)*: when generating data you could want generated values to change in a known interval. Fuzziness allow to specify the maximum delta a generated value can have from the previous value (for the same field), as a delta percentage; value must be between 0.0 and 1.0, where 0 is 0% and 1 is 100%. When not specified there is no constraint on the generated values, boundaries will be defined by the underlying field type
 - `range` *optional (`long` and `double` type only)*: value will be generated between `min` and `max`
-- `cardinality` *optional*: number of different values for the field
+- `cardinality` *optional*: number of different values for the field; note that this value may not be respected if not enough events are generated. Es `cardinality: 1000` with `100` generated events would produce `100` different values, not `1000`.
 - `object_keys` *optional (`object` type only)*: list of field names to generate in a object field type; if not specified a random number of field names will be generated in the object filed type
 - `value` *optional*: hardcoded value to set for the field (any `cardinality` will be ignored)
 - `enum` *optional (`keyword` type only)*: list of strings to randomly chose from a value to set for the field (any `cardinality` will be applied limited to the size of the `enum` values)
--- a/docs/fields-configuration.md
+++ b/docs/fields-configuration.md
@@ -12,7 +12,7 @@ For each config entry the following fields are available:
 - `name` *mandatory*: dotted path field, matching an entry in [Fields definition](./glossary.md#fields-definition)
 - `fuzziness` *optional (`long` and `double` type only)*: when generating data you could want generated values to change in a known interval. Fuzziness allow to specify the maximum delta a generated value can have from the previous value (for the same field), as a delta percentage; value must be between 0.0 and 1.0, where 0 is 0% and 1 is 100%. When not specified there is no constraint on the generated values, boundaries will be defined by the underlying field type
 - `range` *optional (`long` and `double` type only)*: value will be generated between `min` and `max`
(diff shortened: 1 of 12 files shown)


docs/fields-configuration.md
```diff
@@ -12,7 +12,7 @@ For each config entry the following fields are available:
 - `name` *mandatory*: dotted path field, matching an entry in [Fields definition](./glossary.md#fields-definition)
 - `fuzziness` *optional (`long` and `double` type only)*: when generating data you could want generated values to change in a known interval. Fuzziness allow to specify the maximum delta a generated value can have from the previous value (for the same field), as a delta percentage; value must be between 0.0 and 1.0, where 0 is 0% and 1 is 100%. When not specified there is no constraint on the generated values, boundaries will be defined by the underlying field type
 - `range` *optional (`long` and `double` type only)*: value will be generated between `min` and `max`
-- `cardinality` *optional*: distribution of different values for the field, expressed as a ratio between a `numerator` and a `denominator`
+- `cardinality` *optional*: number of different values for the field
```
endorama, line 15: What is going to happen if cardinality is greater than the number of events being created? My understanding is that this PR does not fix this behaviour, so if I specify i.e. `cardinality: 1000` but generate `100` events I will get `cardinality: 100` (which is fair, no complaints here). If my understanding is correct I would just mention this here
```suggestion
- `cardinality` *optional*: number of different values for the field; note that this value may not be respected if not enough events are generated. Es `cardinality: 1000` with `100` generated events would produce `100` different values, not `1000`.
```

endorama (approved):

PR #94, opened 2023-05-17
Tot events instead of tot size

aspacca:
talking with different people consuming the package we found out that is preferred to indicate to total number of events to generate rather than the total size.
there are a couple of reasons mostly:
1. current users have knowledge of how many events they want to generate, and find themselves to figure out to translate this information in terms of size instead
2. internally we try to figure out back the number of events from the size: this is required to provide evenly distributed data in features we will want to add (like providing a field configuration definition of a `period` to generate data for in case of a `date` field): the way we do this is to rendere once the target template and divide its size by the total size requested. this indeed is a limit and it requires a better strategy (like rendering the template multiple times: most probably up to the highest cardinality) in order to achieve a meaningful approximation (currently for `aws.ec2_metrics`, for example, the tool can end up generating double the size requested because of a the non meaningful approximation).

considering the two points above we decided to just drop indicating a total size to generate in favour of a total number of events. beware that it is now possible, omitting the `--tot-events` flag or setting it to `0` to generate an infinite number of events

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -6,9 +6,9 @@ To do this, use the `generate` command. This command targets a specific dataset 
 
 You can pass a local Fields generation configuration file.
 
-`go run main.go generate <package> <dataset> <version> --tot-size <quantity>`
+`go run main.go generate <package> <dataset> <version> --tot-events <quantity>`
 
-`package`, `dataset` and `version` are mandatory. `--tot-size` is mandatory.
+`package`, `dataset` and `version` are mandatory. `--tot-events` is not mandatory and in case it is not provided an infinite number of events will be generated.
 
 **Example**:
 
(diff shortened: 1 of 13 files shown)


docs/usage.md
```diff
@@ -6,9 +6,9 @@ To do this, use the `generate` command. This command targets a specific dataset
 
 You can pass a local Fields generation configuration file.
 
-`go run main.go generate <package> <dataset> <version> --tot-size <quantity>`
+`go run main.go generate <package> <dataset> <version> --tot-events <quantity>`
 
-`package`, `dataset` and `version` are mandatory. `--tot-size` is mandatory.
+`package`, `dataset` and `version` are mandatory. `--tot-events` is not mandatory and in case it is not provided an infinite number of events will be generated.
```
endorama, line 11: What do you think of setting default `tot-events` to `1`? I'm thinking about the principle of least surprise and helping with command discovery for new devs. An infinite stream would be a big surprise to me and not useful if I'm learning the tool or discovering its capabilities.
Creating 1 by default could also help when writing a template, when you generally want 1 event at a time for testing.

endorama (approved):
left a comment on default value for `tot-events` but PR looks good!

PR #95, opened 2023-05-18
Time period

aspacca:
Date time field can now be configured to be evenly generated across a defined period as between `time.Now()` and `time.Now().Add(period)`, where period is `time.Duration`

This way users can simulate events across N hours/days (/or whatever)

### DIFF
--- a/docs/fields-configuration.md
+++ b/docs/fields-configuration.md
@@ -13,6 +13,7 @@ For each config entry the following fields are available:
 - `fuzziness` *optional (`long` and `double` type only)*: when generating data you could want generated values to change in a known interval. Fuzziness allow to specify the maximum delta a generated value can have from the previous value (for the same field), as a delta percentage; value must be between 0.0 and 1.0, where 0 is 0% and 1 is 100%. When not specified there is no constraint on the generated values, boundaries will be defined by the underlying field type
 - `range` *optional (`long` and `double` type only)*: value will be generated between `min` and `max`
 - `cardinality` *optional*: number of different values for the field; note that this value may not be respected if not enough events are generated. Es `cardinality: 1000` with `100` generated events would produce `100` different values, not `1000`.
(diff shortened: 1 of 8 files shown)


pkg/genlib/generator_interface.go
```diff
@@ -22,6 +22,12 @@ import (
 	"time"
 )
 
+var timeNow time.Time
+
+func init() {
```
endorama, line 27: Is there a way to not use `init` here? I would prefer this to be explicit to reduce unexpected side effects.

docs/fields-configuration.md
```diff
@@ -13,6 +13,7 @@ For each config entry the following fields are available:
 - `fuzziness` *optional (`long` and `double` type only)*: when generating data you could want generated values to change in a known interval. Fuzziness allow to specify the maximum delta a generated value can have from the previous value (for the same field), as a delta percentage; value must be between 0.0 and 1.0, where 0 is 0% and 1 is 100%. When not specified there is no constraint on the generated values, boundaries will be defined by the underlying field type
 - `range` *optional (`long` and `double` type only)*: value will be generated between `min` and `max`
 - `cardinality` *optional*: number of different values for the field; note that this value may not be respected if not enough events are generated. Es `cardinality: 1000` with `100` generated events would produce `100` different values, not `1000`.
+- `period` *optional (`date` type only)*: values will be evenly generated between `time.Now()` and `time.Now().Add(period)`, where period is expressed as `time.Duration`
```
gizas, line 16: It would be nice to add an example of usage in one of templates maybe or here in doc for the usage
in configs.yaml
```yaml
- field:
  - name: testperiod
    period: "1h"
```

and in fields.yaml
```yaml
- name: testperiod
  type: date 
```

gizas (approved):

endorama (approved):

PR #98, opened 2023-05-24
Add seed flag

aspacca:
We add a `--seed` flag to feed the rand seed used in pkg and by go-randomdata dependecy.

`timeNow` and `randSeed` are moved to their init functions in pkg instead of passed to the generator factories

`aws.ec2_logs` assets have been refactored so to achieve reproducible generation given the same `--now` and `--seed` value

### DIFF
--- a/pkg/genlib/generator.go
+++ b/pkg/genlib/generator.go
@@ -163,9 +163,22 @@ func generateTemplateFromField(cfg Config, fields Fields, templateEngine int) ([
 	return templateBuffer.Bytes(), objectKeysField
 }
 
-func NewGenerator(cfg Config, flds Fields, totEvents uint64, timeNow time.Time) (Generator, error) {
+func NewGenerator(cfg Config, flds Fields, totEvents uint64) (Generator, error) {
 	template, objectKeysField := generateCustomTemplateFromField(cfg, flds)
 	flds = append(flds, objectKeysField...)
 
-	return NewGeneratorWithCustomTemplate(template, cfg, flds, totEvents, timeNow)
+	return NewGeneratorWithCustomTemplate(template, cfg, flds, totEvents)
+}
+
+// InitGeneratorTimeNow sets base timeNow for `date` field
+func InitGeneratorTimeNow(timeNow time.Time) {
+	// set timeNowToBind to --now flag (already parsed or now)
+	timeNowToBind = timeNow
+}
+
+// InitGeneratorRandSeed sets rand seed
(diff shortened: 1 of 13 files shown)


pkg/genlib/generator.go
```diff
@@ -163,9 +163,22 @@ func generateTemplateFromField(cfg Config, fields Fields, templateEngine int) ([
...
+// InitGeneratorTimeNow sets base timeNow for `date` field
+func InitGeneratorTimeNow(timeNow time.Time) {
+	// set timeNowToBind to --now flag (already parsed or now)
+	timeNowToBind = timeNow
+}
+
+// InitGeneratorRandSeed sets rand seed
+func InitGeneratorRandSeed(randSeed int64) {
+	// set rand and randomdata seed to --seed flag (custom or 1)
+	rand.Seed(randSeed)
```
endorama, line 182: `math/rand.Seed` is [deprecated](https://pkg.go.dev/math/rand#Seed) and it seems the deprecation affects our use case.
I would not rely on the global PRNG and instantiate one with the generator.

Also the `rand.Seed` default behaviour changes in Go 1.20:
> If Seed is not called, the generator is seeded randomly at program startup.
> Prior to Go 1.20, the generator was seeded like Seed(1) at program startup.

Due to this I think we need to ensure to always call Seed with an appropriate default or when using Go 1.20 users of `genlib` library may experience unexpected behaviours.

endorama (changes requested):

pkg/genlib/generator.go
aspacca: > Due to this I think we need to ensure to always call Seed with an appropriate default or when using Go 1.20 users of `genlib` library may experience unexpected behaviours.

that's happening indeed: since `randSeed`, if `--seed` is not passed, has a default value of `1`

I've replaced `rand.Seed` with a `customRand = rand.New()`

endorama: Actually as @aspacca noticed, in Go 1.19.x `rand.Seed` is not deprecated, but in 1.20.0 the method is deprecated **and** the behaviour changes. Also the Go docs is wrong, as noted here: https://stackoverflow.com/questions/75597325/rand-seedseed-is-deprecated-how-to-use-newrandnewseed

It should be `rand.New` not `NewRand`

gizas (approved):

pkg/genlib/generator.go
aspacca: @endorama done!

endorama (approved):

PR #101, opened 2023-06-05
print generator values from TestMain

### LINKED ISSUE #21: Add PRNG seeding for reproducible data generation in testing environment
During unit testing it's necessary to generate data in a producible way. To do so it should be possible to seed the PRNG before the generation step.

aspacca:
closes #21 

we want to have some fuzziness in the tests and not set a fixed seed and timeNow so that on the long run we might be able to catch corner case
but we print them in the tests so in case of failure we can seed the same in order to reproduce.


aspacca:
@endorama ready to merge? thanks

endorama (approved):

PR #107, opened 2023-10-05
[DRAFT]Add generate k8s node data scripts and templates

MichaelKatsoulis:


assets/templates/kubernetes.node/configs.yml
```diff
@@ -0,0 +1,33 @@
+fields:
```
aspacca, line 1: please, move these files in `schema-b` subfolder, thanks :)

generate-node-metrics.sh
```diff
@@ -0,0 +1,120 @@
+function exit_handler()
```
aspacca, line 1: this will be most probably taken care by `elastic-package benchmark rally` (https://github.com/elastic/elastic-package/issues/1475, https://github.com/elastic/elastic-package/issues/984#issuecomment-1733191971)

assets/templates/kubernetes.node/configs.yml
```diff
@@ -0,0 +1,33 @@
...
+    value: 10000
+  - name: event.duration
+    range:
+      min: 1
+      max: 4000000
+  - name: rangeofid
+    range:
+      min: 0
+      max: 10000
+  - name: timedate
```
aspacca, line 26: you can have a date type with `period` ([values will be evenly generated between `time.Now()` and `time.Now().Add(period)`, where `period` is expressed as `time.Duration`](https://github.com/elastic/elastic-integration-corpus-generator-tool/blob/main/docs/fields-configuration.md))

along with `-n, --now date               time to use for generation based on now (date type)` when invoking `elastic-integration-corpus-generator-tool generate-with-template`

MichaelKatsoulis: Do you have any example on how to make this work? I did not find any other usage of `period`. If I set the field as `10s` how can I then use it in the tpl file?

MichaelKatsoulis: Interesting!

MichaelKatsoulis: @aspacca I tried this out and the timestamps created are in the format `2023-10-06 11:42:06.999999 +0000 UTC` which is not [strict_date_optional_time||epoch_millis](https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-date-format.html). They seem to cannot be parsed by ES when trying to upload the generated events.

MichaelKatsoulis: Done in https://github.com/elastic/elastic-integration-corpus-generator-tool/pull/107/commits/6991fc4d50efa475b37c4eec87899986f770efe4

ruflin:
Do we still need this PR? I assume this would not land directly in the k8s integration?

PR #110, opened 2023-10-06
Do not expose GenState

aspacca:
`GenState` was initially owned and initialized by the invoker of the generator library.
More or less since the introduction of the template engines, the state of the generator is owned by the generator instance itself, but still it was passed to the `Generator.Emit()` method that would override the state passed by the invoker.
This is counter intuitive and could lead in the wrong assumption from the generator consumer about the ownership of the state.
This PR removes passing the state to `Generator.Emit()` and makes the state and its factory (`GenState`, `NewGenState()`) private

### DIFF
--- a/internal/corpus/generator.go
+++ b/internal/corpus/generator.go
@@ -129,8 +129,6 @@ func (gc GeneratorCorpus) eventsPayloadFromFields(template []byte, fields Fields
 		return err
 	}
 
-	state := genlib.NewGenState()
-
 	var buf *bytes.Buffer
 	if len(template) == 0 {
 		buf = bytes.NewBuffer(createPayload)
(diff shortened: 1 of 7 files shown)


aspacca:
@scunningham , please review this PR.
the generator tool made quite some progress since v0.3.0 that you introduced in horde
specifically the json generation according to the fields schema is faster now
I'd suggest you to upgrade to the new version we will release (v0.7.0) after merging this PR

another break in the API is that the factory that horde uses (https://github.com/elastic/elastic-integration-corpus-generator-tool/blob/main/pkg/genlib/generator.go#L168) not takes the tot events to generate as part of the parameters, and the state is indeed owned by the generator

scunningham (approved):
Looks good to me.
I no longer maintain the horde stuff, but will cc' the new maintainer.

FYI: @juliaElastic

endorama (approved):

PR #111, opened 2023-10-06
Updating k8s pod and container templates with no host ips and realistic hostnames

gizas:
TSDB nightly tests demand a more relastic set of data generated. As part of this enhancement we update kubernetes pod and container templates with following:

- Removing ips and mac addresses to simulate the fix we introduced in https://github.com/elastic/integrations/issues/6674
- Adding realistic host and node names. Our template need a set of 90 nodes where the agents will be located. We correlate the node names with host names

## How to test:
1. Checkout this PR
2. Produce pod and container samples:

```bash
./elastic-integration-corpus-generator-tool generate-with-template ./assets/templates/kubernetes.pod/gotext.tpl ./assets/templates/kubernetes.pod/fields.yml -c ./assets/templates/kubernetes.pod/configs.yml -y gotext -t 2

./assets/templates/kubernetes.container/gotext.tpl ./assets/templates/kubernetes.container/fields.yml -c ./assets/templates/kubernetes.container/configs.yml -y gotext -t 2
``` 

Check that host.ips and host.mac fields are removed
Check the node names created and have the same value as hostnames.

## Screenshot
Screenshot 2023-10-06 at 1 09 03 PM


assets/templates/kubernetes.pod/gotext_multiline.tpl
```diff
@@ -198,332 +199,8 @@
...
-         "fe80::829:49ff:fee5:3ea5",
-         "10.232.83.1",
-         "fe80::9cc1:ccff:fe21:8cd2",
-         "10.232.83.1",
-         "fe80::94cf:87ff:fe1f:7a92",
-         "10.232.83.1",
-         "fe80::5d:14ff:fec1:2d3a"
-      ],
-      "name": "{{ $agentName }}",
+      "name": "{{- $nodeid := generate "nodeid" -}}",
```
aspacca, line 202: beware, if you invoke `generate "nodeid"` two times in the same template you should end up with two different values

you should: because looking at the screenshot you posted it does not seems so, but this is something to investigate, since it seems to be bug :)

assets/templates/kubernetes.container/gotext_multiline.tpl
```diff
@@ -103,12 +104,12 @@
       },
       "node":{
          "uid": "{{ $uId }}" ,
-         "hostname":"{{ $agentName }}.c.elastic-obs-integrations-dev.internal",
-         "name":"{{ $agentName }}-{{ $rangeofid }}",
+         "hostname":"node-{{ $nodeid }}",
```
martijnvg, line 107: Maybe `host-{{ $nodeid }}`?

assets/templates/kubernetes.container/gotext_multiline.tpl
```diff
@@ -214,7 +215,7 @@
       "dataset":"kubernetes.container"
    },
    "host":{
-      "hostname":"{{ $agentName }}",
+      "hostname":"node-{{ $nodeid }}",
```
martijnvg, line 218: Maybe `host-`?

assets/templates/kubernetes.pod/gotext_multiline.tpl
```diff
@@ -28,12 +29,12 @@
    "kubernetes": {
       "node":{
          "uid": "{{ $uId }}" ,
-         "hostname":"{{ $agentName }}.c.elastic-obs-integrations-dev.internal",
-         "name":"{{ $agentName }}-{{ $rangeofid }}",
+         "hostname":"node-{{ $nodeid }}",
```
martijnvg, line 32: Maybe `host-`?

assets/templates/kubernetes.pod/gotext_multiline.tpl
gizas: ok once per file. Then I refer to nodeid variable

assets/templates/kubernetes.container/gotext_multiline.tpl
gizas: done

aspacca:
@gizas could you please move the files of each folder inside the `schema-b` subfolder? thanks :)

gizas:
> @gizas could you please move the files of each folder inside the `schema-b` subfolder? thanks :)

Done @aspacca

assets/templates/kubernetes.container/schema-b/configs.yml
```diff
@@ -45,7 +45,7 @@ fields:
   - name: container.name
     enum: ["web", "default-http-backend", "dnsmasq", "csi-driver", "web", "web", "web", "prometheus", "konnectivity-agent", "sidecar", "kubedns", "metrics-server-nanny", "web", "web", "fluentbit", "autoscaler", "gke-metrics-agent", "elastic-agent", "web", "kube-state-metrics", "metrics-server", "fluentbit", "elastic-agent", "web", "prometheus-to-sd-exporter"]
   - name: timedate
-    # Add more dates below in case you want your data to be spread
+    # Add more dates below in case you want your data to be spread.Those are fixed because this is required in the nightly tests of Rally
```
aspacca, line 48: can you explain about this?
will it be fixed "forever", or only for the current nightly tests of rally.
what happens if we use `timestamp` with `period: 2880m` (two days) like I did [here](https://github.com/elastic/integrations/pull/8444)

gizas: This one is fixed for the current nightly tests of rally. We have specific visualisations and we want our data to be in specific range in order our queries to be be always specific

aspacca: would it be possible to change the visualisation range to be `-2d/now`?

gizas: @aspacca 
I have made some tests today and from generator tool side we can change the timedate to date type with [period: 48h] and all seem ok. So I dont see an issue with this now

Moreover on the rally side we have hardcoded the values of visualisations [here](https://github.com/elastic/rally-tracks/pull/513)

So we need to find a way to import to calculate the new itervals and end_time based on the track provided. 
cc @martijnvg for visibility

ruflin:
I assume eventually we will move these files over to the k8s package?

gizas:
>I assume eventually we will move these files over to the k8s package?

This is the purpose of [8443](https://github.com/elastic/integrations/pull/8443) and [8444](https://github.com/elastic/integrations/pull/8444)

martijnvg (approved):
👍

PR #116, opened 2023-11-15
Refactor date period last event now

aspacca:
We want to be able to set a `period` for the date field where, instead having the first generated event on now, we will have the last one on now (imagine the case when used with the timestamp of an event).

Since `period` already accepts a negative duration, we achieve this with that: setting a negative duration will results in having the last event generated on now (the case of a negative duration was ignored before, and fall back to no `period` set at all).

There's also another refactoring related to the case of a date field: before each events was generated in span of an hour before now, randomly, with events not in chronological order. This could have required to sort the events afterward.
Now we generated every date field from the date of the previous event (now, for the first event) and a random amount of milliseconds between 0 and 1000


docs/fields-configuration.md
```diff
@@ -13,7 +13,7 @@ For each config entry the following fields are available:
...
 - `range` *optional (`long` and `double` type only)*: value will be generated between `min` and `max`
 - `cardinality` *optional*: number of different values for the field; note that this value may not be respected if not enough events are generated. Es `cardinality: 1000` with `100` generated events would produce `100` different values, not `1000`.
-- `period` *optional (`date` type only)*: values will be evenly generated between `time.Now()` and `time.Now().Add(period)`, where period is expressed as `time.Duration`
+- `period` *optional (`date` type only)*: values will be evenly generated between `time.Now()` and `time.Now().Add(period)`, where period is expressed as `time.Duration`, it accepts also a negative value (in order to have the last event generated on "now")
```
tommyers-elastic, line 16: to clarify - does a negative value of period mean values will be evenly generated between `time.Now().Add(period)` and `time.Now()`? in this case period is negative so the first term is a time in the past.

pkg/genlib/generator_interface.go
```diff
@@ -48,8 +48,8 @@ const (
 	FieldTypeFlattened       = "flattened"
 	FieldTypeGeoPoint        = "geo_point"
 
-	FieldTypeTimeRange  = 3600 // seconds
-	FieldTypeTimeLayout = "2006-01-02T15:04:05.999999Z07:00"
+	FieldTypeDurationSpan = 1000 // milliseconds
```
tommyers-elastic, line 51: can we use typed constants for these to make it clearer e.g. `1000 * time.Millisecond` ?

pkg/genlib/generator_interface.go
```diff
@@ -839,14 +845,20 @@ func bindNearTimeWithReturn(fieldCfg ConfigField, field Field, fieldMap map[stri
 	var emitF EmitF
 	emitF = func(state *genState) any {
 		var offset time.Duration
-		if fieldCfg.Period > 0 {
+		if fieldCfg.Period > 0 && state.totEvents > 0 {
```
tommyers-elastic, line 848: can we avoid the repetition here? (it looks identical to the section above, is it?)

docs/fields-configuration.md
aspacca: yes, it is correct.
I can specify in the docs

pkg/genlib/generator_interface.go
tommyers-elastic: ah - rereading the PR description i understand this better. 

instead of spreading out across an hour, each event is between 0 and 1 s after the previous event. 

this could have a really significant impact on the time span over which events are generated... what's the reason for making this change?

aspacca: > can we use typed constants for these to make it clearer e.g. `1000 * time.Millisecond` ?

I use the constant as param of `rand.Intn`, I cannot use a duration unluckily

pkg/genlib/generator_interface.go
aspacca: do you mean in the other function `bindNearTime` at line 498?
every "bind*" function is repeated, once to write to a buffer, once ("bind*WithReturn") to return a value, according to the template type we use (placeholder vs text go template).

If you think it is worth I can try to extract the common logic where possible and just encapsulate a subcall in the two different version of the "bind*" functions, but I suspect it will have impact on performance, that are a sensible topic for the placeholder template (where we look for the highest throughput possible trading off on template features)

pkg/genlib/generator_interface.go
aspacca: > this could have a really significant impact on the time span over which events are generated... what's the reason for making this change?

yes, it has a significant impact. but the fields for this difference will apply should not be significant.
I mean that if you want a precise time span, you should you `period`. if you are not using period I wanted to achieve to things:
1. no floating back and forth in time the generated dates (always progressive from now)
2. keeping the time span shorter.

I'm open the reconsider my goals :)
either both or one of them only :)

I'm trying to figure out what would be the most "desired" default behavior

tommyers-elastic: on the first part, we can just cast the `Duration` to an `int` in order to pass it to `Intn`?

on the second, sure that sounds fine, but it would be great to see design decisions like this recorded somewhere :)

pkg/genlib/generator_interface.go
tommyers-elastic: yeh, the whole section from 501-515.

the logic here is kinda picky, i would much rather not see it repeated. what are the perf implications? isn't it just the overhead of a function call?

aspacca: > what are the perf implications? isn't it just the overhead of a function call?

yes, but that could me too much enough

the placeholder template is more or less micro-optimised for performance and high throughput

see https://github.com/elastic/elastic-integration-corpus-generator-tool/blob/main/docs/performances.md for context

pkg/genlib/generator_interface.go
aspacca: > it would be great to see design decisions like this recorded somewhere

well. I hoped to have recorded it in the PR description :D

aspacca: > on the first part, we can just cast the `Duration` to an `int` in order to pass it to `Intn`?

but then we could end up with nanoseconds spans because it will be `Intn` of nanoseconds. since we print date with milliseconds resolution we will end up with date with no span

tommyers-elastic: but `Duration`s are already in nanoseconds. we are just scaling up to ms before doing the random selection... i think?

aspacca: > but `Duration`s are already in nanoseconds. we are just scaling up to ms before doing the random selection... i think?

```
 time.Duration(customRand.Intn(1000)) * time.Millisecond
```

is different from
```
 time.Duration(customRand.Intn(1000000000))
```

the first generates from 0 to 1000 milliseconds, ie, it has a duration resolution of milliseconds

the latter generates from 0 to 1000000000 nanoseconds, with a nanoseconds duration resolution that we finally lost when we print at milliseconds resolution :)

pkg/genlib/generator_interface.go
aspacca: > yes, but that could be too much enough

run a benchmark before and after, and in the end it isn't, so I've extracted the logic :)

pkg/genlib/generator_interface.go
aspacca: > the two examples produce equivalent durations with different precisions, right?

exactly, that's the point

`customRand.Intn(1000) * time.Millisecond` will return `event1: 15:10:22.058`, `event2: 15:10:22.189`.
while the second will return `event1: 15:10:22.0580000110`, `event2: 15:10:22.0580024090`.

since when we print the `time.Time` we print milliseconds resolution only, in the second case we end up with the same time

pkg/genlib/generator_interface.go
tommyers-elastic: awesome, thanks! great that we can easily benchmark this code to test things like this

aspacca:
@tommyers-elastic  . all good here? can we merge?

ruflin (commented):
As soon as we have this PR in, it will help to have at least one benchmark available that uses a negative period as an example on how to use it.

tommyers-elastic (approved):
approving with the note that i prefer typed time constants :)

aspacca:
> As soon as we have this PR in, it will help to have at least one benchmark available that uses a negative period as an example on how to use it.

https://github.com/elastic/integrations/pull/8443/commits/99351cfa2ae75e8ec353492d965f2bee491ff18b and https://github.com/elastic/integrations/pull/8444/commits/3f2008ebf99277e0e8ce659c54a361c07f0ac26c

PR #117, opened 2023-11-23
Range from to for date type

aspacca:
We want to be able to generate date fields between a range.
We introduce `range.from` and `range.to` config settings that will be considered only for date fields.
It is possible to specify both `from` and `to` or only one of them.
If both are specified, the dates will be generate between `from` and `to`
If only one is specified, the dates will be generated in the interval given by either `from` or `to` and `time.Now()`
In any case the dates generated will keep progressive order, regardless the interval is negative of positive.

### DIFF
--- a/docs/fields-configuration.md
+++ b/docs/fields-configuration.md
@@ -12,8 +12,9 @@ For each config entry the following fields are available:
 - `name` *mandatory*: dotted path field, matching an entry in [Fields definition](./glossary.md#fields-definition)
 - `fuzziness` *optional (`long` and `double` type only)*: when generating data you could want generated values to change in a known interval. Fuzziness allow to specify the maximum delta a generated value can have from the previous value (for the same field), as a delta percentage; value must be between 0.0 and 1.0, where 0 is 0% and 1 is 100%. When not specified there is no constraint on the generated values, boundaries will be defined by the underlying field type
 - `range` *optional (`long` and `double` type only)*: value will be generated between `min` and `max`
+- `range` *optional (`date` type only)*: value will be generated between `from` and `to`. Only one between `from` and `to` can be set, in this case the dates will be generated between `from`/`to` and `time.Now()`. Progressive order of the generated dates is always assured regardless the interval involving `from`, `to` and `time.Now()` is positive or negative. This setting has preference over `period`. 
 - `cardinality` *optional*: number of different values for the field; note that this value may not be respected if not enough events are generated. Es `cardinality: 1000` with `100` generated events would produce `100` different values, not `1000`.
-- `period` *optional (`date` type only)*: values will be evenly generated between `time.Now()` and `time.Now().Add(period)`, where period is expressed as `time.Duration`. It accepts also a negative duration: in this case  values will be evenly generated between `time.Now().Add(period)` and `time.Now()`.
+- `period` *optional (`date` type only)*: values will be evenly generated between `time.Now()` and `time.Now().Add(period)`, where period is expressed as `time.Duration`. It accepts also a negative duration: in this case  values will be evenly generated between `time.Now().Add(period)` and `time.Now()`. This setting is ignored if a `range` with `from`/`to` is configured.
 - `object_keys` *optional (`object` type only)*: list of field names to generate in a object field type; if not specified a random number of field names will be generated in the object filed type
 - `value` *optional*: hardcoded value to set for the field (any `cardinality` will be ignored)
 - `enum` *optional (`keyword` type only)*: list of strings to randomly chose from a value to set for the field (any `cardinality` will be applied limited to the size of the `enum` values)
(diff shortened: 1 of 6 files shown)


docs/fields-configuration.md
```diff
@@ -12,8 +12,9 @@ For each config entry the following fields are available:
...
 - `cardinality` *optional*: number of different values for the field; note that this value may not be respected if not enough events are generated. Es `cardinality: 1000` with `100` generated events would produce `100` different values, not `1000`.
-- `period` *optional (`date` type only)*: values will be evenly generated between `time.Now()` and `time.Now().Add(period)`, where period is expressed as `time.Duration`. It accepts also a negative duration: in this case  values will be evenly generated between `time.Now().Add(period)` and `time.Now()`.
+- `period` *optional (`date` type only)*: values will be evenly generated between `time.Now()` and `time.Now().Add(period)`, where period is expressed as `time.Duration`. It accepts also a negative duration: in this case  values will be evenly generated between `time.Now().Add(period)` and `time.Now()`. This setting is ignored if a `range` with `from`/`to` is configured.
```
ruflin, line 17: Instead of ignoring it, should we throw an error as this seems to be an invalid config?

docs/fields-configuration.md
```diff
@@ -12,8 +12,9 @@ For each config entry the following fields are available:
 - `name` *mandatory*: dotted path field, matching an entry in [Fields definition](./glossary.md#fields-definition)
 - `fuzziness` *optional (`long` and `double` type only)*: when generating data you could want generated values to change in a known interval. Fuzziness allow to specify the maximum delta a generated value can have from the previous value (for the same field), as a delta percentage; value must be between 0.0 and 1.0, where 0 is 0% and 1 is 100%. When not specified there is no constraint on the generated values, boundaries will be defined by the underlying field type
 - `range` *optional (`long` and `double` type only)*: value will be generated between `min` and `max`
+- `range` *optional (`date` type only)*: value will be generated between `from` and `to`. Only one between `from` and `to` can be set, in this case the dates will be generated between `from`/`to` and `time.Now()`. Progressive order of the generated dates is always assured regardless the interval involving `from`, `to` and `time.Now()` is positive or negative. This setting has preference over `period`.
```
ruflin, line 15:
```suggestion
- `range` *optional (`date` type only)*: value will be generated between `from` and `to`. Only one between `from` and `to` can be set, in this case the dates will be generated between `from`/`to` and `time.Now()`. Progressive order of the generated dates is always assured regardless the interval involving `from`, `to` and `time.Now()` is positive or negative. This setting has preference over `period`. The format of the date is as following: `"2023-11-23-11:29:48Z"`
```

ruflin (commented):
@gizas Can you please take a look at this and see if it solves your problem of generating a corpus for a very specific date range?

gizas:
@aspacca thanks for this. Really helpful

I have made this draft branch to test  [here](https://github.com/elastic/elastic-integration-corpus-generator-tool/blob/timedateperiod/assets/templates/kubernetes.pod/configs.yml#L47-L50)

Some minor comments:
- The from/to dates must have this format `from: "2023-05-15T00:00:00.999999999-00:01"` otherwise we have parse error `Error: parsing time "2023-05-15" as "2006-01-02T15:04:05.999999999-07:00": cannot parse "" as "T" accessing 'fields.15.range.from'`. 
  I dont mind, but please change the documentation example here: https://github.com/elastic/elastic-integration-corpus-generator-tool/pull/117/files#diff-3e822219875fd3448087df4174dd709f82eb3b229fc3a64cddcea005e4eb2f2dR32-R33

- In order to make the date compatible with what we have in ES , I had to format it again like here: https://github.com/elastic/elastic-integration-corpus-generator-tool/blob/timedateperiod/assets/templates/kubernetes.pod/gotext.tpl#L3
Mind that in your format is `999999999-00:01` and in previous templates we had `999999Z07+00:00`. So we revert `-` with `+`. Do you think shall we also change this?

- Last one (minor)
See a sample of results:
```bash
2023-05-15T00:00:00.999999-00+00:00"
"2023-05-15T01:12:00.974999-00+00:00"
"2023-05-15T02:24:00.949999-00+00:00"
"2023-05-15T03:36:00.924999-00+00:00"
"2023-05-15T04:48:00.899999-00+00:00"
"2023-05-15T06:00:00.874999-00+00:00"
"2023-05-15T07:12:00.849999-00+00:00"
```
They are sorted (really cool !!!), only thing is that last part is always `-00+00:00`. Maybe we dont need such accuracy but worth to mentioned it

gizas:
Thanks Andrew.

So I think you can skip the micro seconds for now.

I would only suggest the documentation update for the correct timestamp and also to update this doc with a reminder/mention to an example for the format.

aspacca:
@gizas docs fixed!

gizas:
Andrew still the eg. fails for parsing
`Error: parsing time "2023-11-23T11:29:48" as "2006-01-02T15:04:05.999999999-07:00": cannot parse "" as "-07:00" accessing 'fields.15.range.from'`

Please can you update the example https://github.com/elastic/elastic-integration-corpus-generator-tool/pull/117/files#diff-3e822219875fd3448087df4174dd709f82eb3b229fc3a64cddcea005e4eb2f2dR32-R33

aspacca:
> Andrew still the eg. fails for parsing

I should have tested it! 😭

aspacca:
@gizas https://go.dev/play/p/V5COIdqSILz

Issue #124, opened 2023-12-27
Support generating counter values


ali786XI:
As of now, the `corpus-generator-tool` produces different `gauge` values with the use of `ranges`, `enum` and `fuzziness` as per the config [documentation](https://github.com/elastic/elastic-integration-corpus-generator-tool/blob/main/docs/fields-configuration.md).

However there can be a config defining a particular field as `counter` so it always produces incremental values. There are many such packages which are TSDB migrated which are using these `counter` [fields](https://github.com/elastic/integrations/blob/main/packages/apache_spark/data_stream/driver/fields/fields.yml#L24). It would be easier to implement the `rally` benchmarking templates as well.

- [ ] Counter support with no reset ([PR](https://github.com/elastic/elastic-integration-corpus-generator-tool/pull/129))
- [ ] Support for counter reset

Counters can be reset:
1. having the possibility to enable/disable just totally random occurrences of the reset.
2. having the possibility to define the probability that the counter is reset every time a new value is generated
3. having the possibility to specify the number of events after which the counter must be reset

Proposal for a `counter_reset` configuration.
```
1. 
counter_reset: 
  strategy: random
2.
counter_reset: 
  strategy: probability
  probability: 0-100 # %
3.
counter_reset: 
  strategy: events_amount
  events_amount: N
```

cc:- @aspacca

ruflin:
++ on supporting counters. @aspacca How much effort would it be to add this?

aspacca:
@ruflin code wise the effort is minimal, I have to come up with some ideas on how to configure the feature in the config file :)

ruflin:
@aspacca Can you share a bit more on your thoughts and put them down here for further discussion?

aspacca:
@ruflin 

on the code side is pretty straightforward
the current behavior is the following: assuming the previous generated value is `20`, with a fuzziness of `0.6`, the next value `X` generated applying fuzziness is in the range `20 * (1 - 0.6) < X < 20 * (1 + 0.6)`. `range.min` and `range.max` are applied as lowest and highest bounds of the range, if present.

this work wells for gauge

for counter,  `X` must be generated in `20 < X < 20 * (1 + 0.6)`.

on the config side, I'm keen to exclude the possibility to have a field with both `range` and `is_counter` (or something like that). the reason is because if we have a `range`, we'll hit `range.max` after a number big enough of events generated.

[...]

PR #129, opened 2024-01-30
Counter support

aspacca:
Adding `counter` config for `long` and `double` type.
With `counter: true`, ever-increasing values will be generated.
`counter: true` and `range.min/max` config cannot be set at the same time and will result in an error stopping the generator.
`fuzziness` is applied if defined.

Related to https://github.com/elastic/elastic-integration-corpus-generator-tool/issues/124


aspacca:
@ruflin , @aliabbas-elastic . counters can be reset. the case is not covered in this PR.

I see a few alternatives for that:
1. having the possibility to enable/disable just totally random occurrences of the reset.
2. having the possibility to define the probability that the counter is reset every time a new value is generated
3. having the possibility to specify the number of events after which the counter must be reset

If 1. is enough to start from it's fine for me to add directly in this PR.

If and when we will add at least one of the three I suggest to change the `counter` config type to a string. The string value will drive all three cases.

If we decided to add in a separated PR this will result in a breaking change. If we do not want to have a breaking change, we must add instead another separated config to drive the reset cases.

Just giving context in order for you to decide :)


PS: indeed adding `cardinality` to a field with a `counter: true` is the equivalent of 3. This is a the side effect of the behaviour of the current `cardinality`  implementation. We already planned to change this behaviour. Once we implement the new behaviour, `counter: true` and `cardinality` config cannot be set at the same time and will result in an error stopping the generator.

aspacca:
@ruflin , @aliabbas-elastic , sorry, after some considerations I made there is no value in having `counter` also configuring the reset, it is indeed way better to have a separated config.

this excludes at all dealing with a breaking change, so there's no benefit/tradeoff to consider to cover reset in this PR.

I will add to https://github.com/elastic/elastic-integration-corpus-generator-tool/issues/124 a proposal for a reset configuration, covering all the 3 cases mentioned.

docs/fields-configuration.md
```diff
@@ -11,10 +11,11 @@ The config file is a yaml file consisting of root level `fields` object that's a
 For each config entry the following fields are available:
 - `name` *mandatory*: dotted path field, matching an entry in [Fields definition](./glossary.md#fields-definition)
 - `fuzziness` *optional (`long` and `double` type only)*: when generating data you could want generated values to change in a known interval. Fuzziness allow to specify the maximum delta a generated value can have from the previous value (for the same field), as a delta percentage; value must be between 0.0 and 1.0, where 0 is 0% and 1 is 100%. When not specified there is no constraint on the generated values, boundaries will be defined by the underlying field type
-- `range` *optional (`long` and `double` type only)*: value will be generated between `min` and `max`
+- `range` *optional (`long` and `double` type only)*: value will be generated between `min` and `max`, eventually according to the defined `fuzziness`.
```
ruflin, line 14: It is not clear to me what fuzziness means in this context?

docs/fields-configuration.md
```diff
@@ -11,10 +11,11 @@ The config file is a yaml file consisting of root level `fields` object that's a
...
-- `period` *optional (`date` type only)*: values will be evenly generated between `time.Now()` and `time.Now().Add(period)`, where period is expressed as `time.Duration`. It accepts also a negative duration: in this case  values will be evenly generated between `time.Now().Add(period)` and `time.Now()`. If both `period` and at least one of `from` or `to` settings are defined an error will be returned and the generator will stop.
+- `cardinality` *optional*: exact number of different values to generate for the field; note that this setting may not be respected if not enough events are generated. Es `cardinality: 1000` with `100` generated events would produce `100` different values, not `1000`. Similarly, the setting may not be respected if other settings prevents it. Es `cardinality: 10` with an `enum` list of only 5 strings would produce `5` different values, not `10`. Or `cardinality: 10` for a `long` with `range.min: 1` and `range.max: 5` would produce `5` different values, not `10`.
```
ruflin, line 16: > Es `cardinality: 10`

Should this be "Is `cardinality: 10`"?

docs/fields-configuration.md
```diff
@@ -11,10 +11,11 @@ The config file is a yaml file consisting of root level `fields` object that's a
...
+- `cardinality` *optional*: exact number of different values to generate for the field; note that this setting may not be respected if not enough events are generated. Es `cardinality: 1000` with `100` generated events would produce `100` different values, not `1000`. Similarly, the setting may not be respected if other settings prevents it. Es `cardinality: 10` with an `enum` list of only 5 strings would produce `5` different values, not `10`. Or `cardinality: 10` for a `long` with `range.min: 1` and `range.max: 5` would produce `5` different values, not `10`. 
+- `counter` *optional (`long` and  `double` type only)*: if set to `true` values will be generated only ever-increasing, eventually according to the defined `fuzziness`. If both `counter: true` and at least one of `range.min` or `range.max` settings are defined an error will be returned and the generator will stop.
```
ruflin, line 17: Same as above, how does fuzziness play into this. Can you explain a bit more in the description. Maybe use an example like you have above for cardinality.

pkg/genlib/generator_interface.go
```diff
@@ -567,7 +567,48 @@ func fuzzyInt(previous int64, fuzziness, min, max float64) int64 {
 	return customRand.Int63n(int64(math.Ceil(higherBound-lowerBound))) + int64(lowerBound)
 }
 
+func fuzzyIntCounter(previous int64, fuzziness float64) int64 {
+	lowerBound := float64(previous)
+	higherBound := float64(previous) * (1 + fuzziness)
```
ruflin, line 572: fuzziness is to have larger increments?

pkg/genlib/generator_interface.go
```diff
@@ -567,7 +567,48 @@ func fuzzyInt(previous int64, fuzziness, min, max float64) int64 {
...
+				dummyFunc = makeIntCounterFunc(previous, field)
+
+				dummyInt = dummyFunc()
+			} else {
+				dummyInt = fuzzyIntCounter(previous, fieldCfg.Fuzziness)
+			}
+
+			state.prevCache[field.Name] = dummyInt
+			v := make([]byte, 0, 32)
+			v = strconv.AppendInt(v, dummyInt, 10)
```
ruflin, line 602: I must confess, this code looks quite complicated with the buffer and strconv but it seems a pattern that is already used.

pkg/genlib/generator_interface.go
```diff
@@ -617,7 +658,46 @@ func fuzzyFloat(previous, fuzziness, min, max float64) float64 {
...
+			if fieldCfg.Fuzziness <= 0 {
+				dummyFunc = makeFloatCounterFunc(previous, field)
+
+				dummyFloat = dummyFunc()
+			} else {
+				dummyFloat = fuzzyFloatCounter(previous, fieldCfg.Fuzziness)
+			}
+
+			state.prevCache[field.Name] = dummyFloat
+			_, err := fmt.Fprintf(buf, "%f", dummyFloat)
```
ruflin, line 692: Is this to catch potential conversion errors?

pkg/genlib/generator_interface.go
```diff
@@ -934,10 +1114,11 @@ func bindDoubleWithReturn(fieldCfg ConfigField, field Field, fieldMap map[string
 		return nil
 	}
 
+	var emitF emitF
```
ruflin, line 1117: What is the reason this was moved?

ruflin:
* For the reset, agree it should be a separate config. But I would skip it for now. I assume still we have a reset, assuming we reach the max value so the counter starts from 0 again (will take a while to get there with the generator)
* I am a bit confused how fuzziness is used (and if it should be used in the first place). Can you share a bit more thoughts on the usage of fuzziness?
* Lets assume in the template I have 3 counters. If I just enable counter, I assume all start from 0 and with each iteration there is an increment of 1. It means, these 3 counters will always have the exact same value.If this is the case, the side effect could be that ES can better compress these values. I think for now this is ok. Later on, we can discuss if we should be able to define different start values for a counter (or make it random) so that not all counters are aligned.

docs/fields-configuration.md
aspacca: I will add an extra sentence.
Basically it means that every random value will be generated between min and max.
A specific value will be generated within a delta defined by fuzziness from the previous value, if fuzziness is defined. In any case (fuzziness or not) the value would not escape the min/max range.

docs/fields-configuration.md
aspacca: no, sorry :)
It should be `For example`

docs/fields-configuration.md
aspacca: Yes, I will add

pkg/genlib/generator_interface.go
aspacca: fuzziness allows to specify the maximum delta a generated value can have from the previous value.
If you have an int and you don't have fuzziness, you can have a series of generated value like: "12, 1765, 123, 87610"

The same if you have a range. Imagine a cpu usage field, 0-100: "2, 98, 37, 12, 78"

If you don't want those "jumps", but rather something smoother, you rely on fuzziness.

For the cpu usage field a series with 20% delta is probably more realistic:
```
- name: cpu
  range:
    min: 0
    max: 100
  fuzziness: 0.2
```

the series will look more like: "10, 12, 9, 7"

pkg/genlib/generator_interface.go
aspacca: The same for me :)
It was introduced with the huge performance refactoring at https://github.com/elastic/elastic-integration-corpus-generator-tool/pull/15

pkg/genlib/generator_interface.go
aspacca: https://pkg.go.dev/fmt#Fprintf

> err return values from Fprintf are those returned by the underlying io.Writer.

pkg/genlib/generator_interface.go
aspacca: no specific reason, it happened while working on the file. I will revert :)

docs/fields-configuration.md
```diff
@@ -10,11 +10,12 @@ The config file is a yaml file consisting of root level `fields` object that's a
...
+- `counter` *optional (`long` and  `double` type only)*: if set to `true` values will be generated only ever-increasing. If `fuzziness` is defined, the value will be generated within a positive delta defined by `fuzziness` from the previous value. For example, `fuzziness: 0.1`, assuming `counter: true` , assuming a `double` field type and with first value generated `10.`, will generate the second value in the range between `10.` and `11.`. Assuming the second value generated will be `10.5`, the third one will be generated in the range between `10.5` and `11.55`, and so on. If both `counter: true` and at least one of `range.min` or `range.max` settings are defined an error will be returned and the generator will stop.
```
lalit-satapathy, line 17: > If `fuzziness` is defined

Can we define the behaviour of counter if `fuzziness` is not defined. While theoretically it makes sense, I am wondering which case it will be useful, if counter delta is very high? or should we always ensure `fuzziness` is defined with counter?

lalit-satapathy:
>I like the idea of counter reset but it is not clear to me if it is a top priority. Is there currently an issue we have if no reset happens?
> For the reset, agree it should be a separate config. But I would skip it for now.  

Agree this can be taken up later, I am not sure, where we need such a case immediately.

aliabbas-elastic (approved):
LGTM

PR #180, opened 2026-03-31
fix: long/double enum type assertion failure in gotext templates

### LINKED ISSUE #179: enum with long and double fields fails when using gotext templates
When using `enum` with `long` or `double` typed fields in a gotext template (`-y gotext`), the generator fails. With a single enum field it returns `generate called on a field not present in fields yaml definition`. With two or more enum fields it panics with `close of closed channel`.

This was introduced in #144 which added enum support for these types. The same `enum` configuration works correctly with placeholder templates (`-y placeholder`).

### Steps to reproduce

**fields.yml:**
```yaml
- name: ip_protocol
  type: long
- name: sample_interval
  type: long
```

**config.yml:**
```yaml
fields:
  - name: ip_protocol
    enum: [1, 6, 17, 47]
  - name: sample_interval
    enum: [100, 1000, 10000]
```

**template.ndjson:**
```
{{- $ip_protocol := generate "ip_protocol" -}}
{{- $sample_interval := generate "sample_interval" -}}
{"IPProtocol":{{ $ip_protocol }},"SampleInterval":{{ $sample_interval }}}
```

Run:
```bash
go run main.go generate-with-template template.ndjson fields.yml \
  -c config.yml -y gotext -t 5
```

### Expected behavior

Generates 5 events with values randomly picked from the enum lists.

### Actual behavior

```
Error: template: generator:3:40: executing "generator" at <generate "sample_interval">:
  error calling generate: close of closed channel
exit status 1
```

brijesh-elastic:
## Summary

Fixes a runtime panic when using `enum` with `long` or `double` fields in gotext templates (`-y gotext`).

Closes #179 

## Fix

Use explicit named type declaration:

```go
var emitF emitF
emitF = func(state *genState) any { ... }
```

Applied to both `bindLongWithReturn` and `bindDoubleWithReturn`.

### DIFF
--- a/pkg/genlib/generator_interface.go
+++ b/pkg/genlib/generator_interface.go
@@ -1064,7 +1064,8 @@ func bindLongWithReturn(fieldCfg ConfigField, field Field, fieldMap map[string]a
 	}
 
 	if len(fieldCfg.Enum) > 0 {
-		emitF := func(state *genState) any {
+		var emitF emitF
+		emitF = func(state *genState) any {
 			idx := state.rand.Intn(len(fieldCfg.Enum))
 			f, _ := strconv.ParseInt(fieldCfg.Enum[idx], 10, 64)
 			return f
@@ -1223,7 +1224,8 @@ func bindDoubleWithReturn(fieldCfg ConfigField, field Field, fieldMap map[string
 	}
 
 	if len(fieldCfg.Enum) > 0 {
-		emitF := func(state *genState) any {
+		var emitF emitF
+		emitF = func(state *genState) any {
 			idx := state.rand.Intn(len(fieldCfg.Enum))
 			f, _ := strconv.ParseFloat(fieldCfg.Enum[idx], 64)
 			return f


shmsr (approved):

cavokz:
@brijesh-elastic, @shmsr I truly don't understand how this change can make any difference. Could you please elaborate?

shmsr:
> @brijesh-elastic, @shmsr I truly don't understand how this change can make any difference. Could you please elaborate?

Simple example to explain

```go
package main
import "fmt"
type emitF func() any
func main() {
	broken := map[string]any{
		"alpha": func() any { return int64(42) }, // type is func() any
	}
	fixed := map[string]any{}
	var f emitF
	f = func() any { return int64(42) } // type is emitF
	fixed["alpha"] = f
	_, brokenOK := broken["alpha"].(emitF)
	_, fixedOK := fixed["alpha"].(emitF)
	fmt.Println("broken:", brokenOK)
	fmt.Println("fixed:", fixedOK)
}
```

- emitF is a named function type.
- func() any is an unnamed function type.
- even though their signatures look identical, Go treats them as different runtime types.
- So if you store func() any in map[string]any, value.(emitF) fails.

cavokz:
> Simple example to explain
> 
> ```go
> package main
> import "fmt"
> type emitF func() any
> func main() {
> 	broken := map[string]any{
> 		"alpha": func() any { return int64(42) }, // type is func() any
> 	}
> 	fixed := map[string]any{}
> 	var f emitF
> 	f = func() any { return int64(42) } // type is emitF
> 	fixed["alpha"] = f
> 	_, brokenOK := broken["alpha"].(emitF)
> 	_, fixedOK := fixed["alpha"].(emitF)
> 	fmt.Println("broken:", brokenOK)
> 	fmt.Println("fixed:", fixedOK)
> }
> ```
> 
>     * emitF is a named function type.
> 
>     * func() any is an unnamed function type.
> 
>     * even though their signatures look identical, Go treats them as different runtime types.
> 
>     * So if you store func() any in map[string]any, value.(emitF) fails.

Now it makes a lot of sense, thanks. I wondered so many times why `var f emitF` was scattered all over the places.

Apparently we don't have any unit test that catches this subtleness.

PR #182, opened 2026-06-02
fix: prevent data race in NewGenerator on shared fields slice backing array

cavokz:
## Problem

When multiple goroutines call `NewGenerator` with the **same** `Fields` slice (a common pattern in Horde's enrollment loop), the line:

```go
fields = append(fields, objectKeysField...)
```

inside `newGeneratorWithCustomTemplate` can extend the caller's backing array **in-place** when `cap > len`. This condition arises frequently because `normaliseFields` allocates `cap = len(input)` but may keep fewer entries after filtering wildcard fields.

Two goroutines writing to the same backing-array slots corrupt each other's `objectKeysField` entries. A goroutine then binds a field name that differs from what it put in the template, leaving `fieldMap` without an entry for that token. The subsequent type assertion:

```go
emitFunc: fieldMap[fieldName].(emitFNotReturn),  // panics when nil
```

panics with:

```
interface conversion: interface {} is nil, not genlib.emitFNotReturn
```

The panic is intermittent and timing-dependent (hence hard to reproduce with sequential seeds), but triggers reliably at scale (10K drones × 50 streams).

## Fix

Use a three-index slice expression before the append:

```go
fields = append(fields[:len(fields):len(fields)], objectKeysField...)
```

Setting `cap == len` on the slice header forces `append` to allocate a fresh backing array. The caller's memory is never mutated, so concurrent goroutines sharing the same `Fields` slice are safe.

## Tests added

| Test | What it covers |
|------|---------------|
| `TestNewGenerator_ConcurrentSharedSlice_NoPanic` | 32 goroutines × 200 calls sharing a `Fields` slice with explicit excess capacity (`make(Fields, 0, N+10)`). **Detected the data race with `-race` before this fix; passes cleanly after.** |
| `TestNewGenerator_FlattenedWithSubFields_NoPanic` | Serial sweep of 10 000 seeds with the `auditd_manager/auditd` v1.20.0 flattened field set. |
| `TestNewGenerator_FlattenedWithSubFields_EmitNoPanic` | Same field set, also exercises `Emit`. |
| `TestNewGenerator_RealEPRFields_NoPanic` | Loads the actual EPR YAML (`testdata/auditd_manager_auditd_v1.20.0_fields.yml`) and sweeps 10 000 seeds. |

Testdata added: `auditd_manager_auditd_v1.20.0_fields.yml` (773 lines, 262 fields) and `auditd_manager_auditd_v1.20.0_base-fields.yml` from the real EPR package.


shmsr (commented):
Reviewed and verified the root cause end-to-end: on this branch, reverting the one-line change reliably reproduces both `WARNING: DATA RACE` and the exact `interface conversion: interface {} is nil, not genlib.emitFNotReturn` panic; with the fix it passes. The diagnosis is accurate and `append(fields[:len(fields):len(fields)], objectKeysField...)` is the correct, idiomatic fix — and it's the right place to fix it, because it protects regardless of where the slice originated (including hand-built slices that never pass through `normaliseFields`). This is the canonical fix of the three. 👍

Concerns, all about the tests rather than the fix:

1. **Three of the four tests don't exercise the bug.** `TestNewGenerator_FlattenedWithSubFields_NoPanic` (10k seeds), `…_EmitNoPanic` (1k), and `…_RealEPRFields_NoPanic` (10k) are all **serial**. The panic only manifests under concurrency, so these pass on unfixed `main` too — they're fuzz tests, not regression tests for this bug. Only `TestNewGenerator_ConcurrentSharedSlice_NoPanic` actually guards the fix. Suggest dropping the serial sweeps (or cutting them to a handful of seeds); under CI's `go test -v -shuffle=on` the per-seed `t.Run` produces 20,000+ subtest lines and ~5.5s runtime.

2. **~800 lines of testdata YAML** (`auditd_manager_auditd_v1.20.0_fields.yml` + `base-fields.yml`) exist solely to feed `RealEPRFields_NoPanic` — the serial test that doesn't test the bug. Suggest removing both unless there's a separate reason to keep golden EPR fixtures.

3. **CI doesn't run `-race`** (`ci.yml`: `go test -v -shuffle=on ./...`). Good news: I confirmed `TestNewGenerator_ConcurrentSharedSlice_NoPanic` catches the *panic* reliably even without `-race` (5/5 on unfixed code), so it's a valid guard as-is. But it won't catch the underlying data race in CI — worth a follow-up `-race` job.

4. **`WithRandSeed(0)` is non-deterministic** — seed `0` is treated as "unset", so the `seed := 0` iteration of these loops isn't reproducible (verified: two runs of seed 0 differ; seeds ≥1 are stable). Harmless for a panic/fuzz test, but start the loops at `1` to avoid a misleading "reproducible seed".

5. Minor: `seed := seed` / `g := g` loop-var copies are redundant on Go 1.22.

**On the PR split:** I separately verified that #183 *alone* also clears this race for the production path — so #182 and #183 are two layers fixing the *same* bug, not independent issues. Recommend landing the race fix as **one PR** (this one + the concurrent test, optionally folding in #183's trim as belt-and-suspenders) and keeping #184 as a separate cleanup. See my note on #183.

shmsr (approved):

PR #183, opened 2026-06-02
fix: trim excess capacity from normaliseFields return value

cavokz:
## Problem

`normaliseFields` allocates its output slice with `cap = len(input)`:

```go
normalisedFields := make(Fields, 0, len(fields))
```

When wildcard fields are filtered out (because matching concrete fields exist), the returned slice has `len < cap`. The surplus capacity is a latent data-race vector: any caller that shares this slice across goroutines and later appends to it can write into backing-array slots that other goroutines are concurrently reading or writing, without Go's runtime detecting a conflict at the slice-header level.

This is the underlying condition that enables the data race fixed in `newGeneratorWithCustomTemplate` — but it affects **all callers** of `normaliseFields`, not just that function.

## Fix

Add a three-index slice expression before returning to trim the excess:

```go
normalisedFields = normalisedFields[:len(normalisedFields):len(normalisedFields)]
return normalisedFields, nil
```

With `cap == len`, any subsequent `append` by a caller **must** allocate a fresh backing array and cannot interfere with other users of the same slice. This is a standard Go idiom for returning a slice that is safe to append to from multiple goroutines without additional synchronisation.

### DIFF
--- a/pkg/genlib/fields/fields.go
+++ b/pkg/genlib/fields/fields.go
@@ -81,5 +81,11 @@ func normaliseFields(fields Fields) (Fields, error) {
 	}
 
 	sort.Sort(normalisedFields)
+	// Trim excess capacity so that callers sharing this slice across goroutines
+	// cannot inadvertently share the backing array when they append to it.
+	// normaliseFields allocates cap=len(fields) but may keep fewer entries when
+	// wildcard fields are filtered; without this trim the surplus capacity is
+	// a latent data-race vector (see fbxj-vtfy-yrbl-mlyw).
+	normalisedFields = normalisedFields[:len(normalisedFields):len(normalisedFields)]
 	return normalisedFields, nil
 }


shmsr (commented):
Correct and harmless — and I empirically confirmed it does what it claims: `LoadFieldsWithTemplateFromString` on a field set with filtered wildcards now returns `cap == len` (spare 0, vs spare 30 on `main`), and a concurrent reproducer sharing the `normaliseFields` output across 32 goroutines panics ~4.5k times on `main` but passes cleanly here — **both plain and under `-race`**.

That verification is exactly the key point for the PR structure:

1. **#183 alone fixes the production-path race — and so does #182 alone.** They are two layers of defense for the *same* panic, not independent fixes (only #184 is genuinely independent, being a no-op on output). The "they are independent" framing in the split is misleading for these two. #182 is the more robust layer because it protects any slice; #183 only protects slices that flow through `normaliseFields`.

   Recommendation: land the race fix as **one PR** (#182 as canonical + the concurrent regression test, with this trim folded in as belt-and-suspenders), and keep #184 separate. As-is, three PRs over-splits a single fix and risks a reviewer merging only one of #182/#183. If you keep this separate, please reword the description so it reads as hardening for the same bug, and note the narrower scope.

2. **The "safe to share across goroutines" framing is too strong.** `cap == len` only forces a *future append* to reallocate; it does nothing for concurrent in-place element mutation. Suggest narrowing the wording to "safe to append to".

3. **No test coverage** — `pkg/genlib/fields` has no test files. A 2-line unit test asserting `cap(out) == len(out)` would lock this in and document intent.

shmsr (approved):

PR #184, opened 2026-06-02
fix: resolve leaf type for flattened/object sub-key fields in template generation

cavokz:
## Problem

In `generateTemplateFromField`, when a `flattened`, `object`, or `nested` field generates random sub-key entries (`objectKeysField`), each sub-key was built by mutating the parent field in-place:

```go
originalFieldName := field.Name
field.Name = fieldNameRoot + "." + rNoun
objectKeysField = append(objectKeysField, field)
field.Name = originalFieldName
```

Each appended entry therefore **inherits the parent's `Type`** (e.g. `"flattened"`). When `bindField` processes such an entry it routes through `bindObject → bindDynamicObject` instead of `bindKeyword` directly. `bindObject` resets the type to `keyword` before delegating, so the final emitter is correct — but the extra indirection is unnecessary, obscures intent, and creates an avoidable wrapper layer around the emitter.

There is also a subtle mutation hazard: if `field` is later referenced again in the same loop iteration after the restore (`field.Name = originalFieldName`), the temporary name change could in principle produce unexpected side effects.

## Fix

Copy the field into a new variable and resolve the leaf type before appending:

```go
subKeyField := field
subKeyField.Name = fieldNameRoot + "." + rNoun
if subKeyField.ObjectType != "" {
    subKeyField.Type = subKeyField.ObjectType
} else {
    subKeyField.Type = FieldTypeKeyword
}
objectKeysField = append(objectKeysField, subKeyField)
```

- The original `field` is no longer mutated.
- Sub-keys receive the correct leaf type so `bindField` calls `bindKeyword` directly, matching the semantic intent.

### DIFF
--- a/pkg/genlib/generator.go
+++ b/pkg/genlib/generator.go
@@ -129,10 +129,8 @@ func generateTemplateFromField(cfg Config, fields Fields, templateEngine int, st
 					}
 				}
 
-				// Build the sub-key field with the resolved leaf type so that
-				// bindField uses bindKeyword directly rather than recursing
-				// through bindObject → bindDynamicObject with Type = "flattened"
-				// (which could wrap a nil emitter under race conditions).
+				// Resolve the leaf type so bindField binds the sub-key directly
+				// (e.g. keyword/long) instead of re-entering bindObject → bindDynamicObject.
 				subKeyField := field
 				subKeyField.Name = fieldNameRoot + "." + rNoun
 				if subKeyField.ObjectType != "" {
--- a/pkg/genlib/generator.go
+++ b/pkg/genlib/generator.go
@@ -129,10 +129,18 @@ func generateTemplateFromField(cfg Config, fields Fields, templateEngine int, st
 					}
 				}
 
-				originalFieldName := field.Name
-				field.Name = fieldNameRoot + "." + rNoun
-				objectKeysField = append(objectKeysField, field)
-				field.Name = originalFieldName
+				// Build the sub-key field with the resolved leaf type so that
+				// bindField uses bindKeyword directly rather than recursing
+				// through bindObject → bindDynamicObject with Type = "flattened"
+				// (which could wrap a nil emitter under race conditions).
+				subKeyField := field
+				subKeyField.Name = fieldNameRoot + "." + rNoun
+				if subKeyField.ObjectType != "" {
+					subKeyField.Type = subKeyField.ObjectType
+				} else {
+					subKeyField.Type = FieldTypeKeyword
+				}
+				objectKeysField = append(objectKeysField, subKeyField)
 
 				templateBuffer.WriteString(fieldTemplate)
 			}


shmsr (commented):
I traced both binding paths and then verified the result empirically.

- **Old:** sub-key keeps `Type="flattened"` → `bindField` → `bindObject` (resets type to keyword) → `bindDynamicObject` → wraps the keyword emitter in `makeDynamicStub`.
- **New:** sub-key gets `Type=keyword` up front → `bindKeyword` directly.

`makeDynamicStub` just pipes the inner emitter's bytes through a pooled buffer (emitting nothing only if the inner emits nothing, which keyword never does). I confirmed with deterministic seeds (1–8) across `flattened`/`nested`/`object(ObjectType=long)`/`keyword` fields that the emitted output is **byte-identical** before and after this change. (Caveat: `WithRandSeed(0)` is non-deterministic, so seed 0 must be excluded from any equivalence diff — seeds ≥1 are stable.)

So this is a safe, behavior-preserving cleanup — it removes a needless wrapper plus the in-place `field.Name` mutate/restore, and it's genuinely independent of the race fix (can land on its own).

**One thing to fix before merge — the code comment is inaccurate:**

> `// (which could wrap a nil emitter under race conditions).`

It couldn't. The nil under the race was in the `fieldMap[fieldName]` lookup in the emitters loop (`generator_with_custom_template.go:116`), entirely outside this code path. This PR would **not** have prevented the panic — #182 does. Suggest reframing the PR description and the comment as a clarity/cleanup change ("bind sub-keys as their resolved leaf type directly"), with no race-fix implication.

shmsr (commented):
.

shmsr (commented):
Please update the comment — this path doesn’t wrap a nil emitter under races (that’s #182):

```go
// (which could wrap a nil emitter under race conditions).
```

Reframe it as resolving the sub-key leaf type so bindField binds it directly, with no race implication.

shmsr:
Suggested replacement for the inaccurate race comment:

```go
// Resolve the leaf type so bindField binds the sub-key directly
// (e.g. keyword/long) instead of re-entering bindObject → bindDynamicObject.
```

pkg/genlib/generator.go
```diff
@@ -129,10 +129,18 @@ func generateTemplateFromField(cfg Config, fields Fields, templateEngine int, st
...
 				}
 
-				originalFieldName := field.Name
-				field.Name = fieldNameRoot + "." + rNoun
-				objectKeysField = append(objectKeysField, field)
-				field.Name = originalFieldName
+				// Build the sub-key field with the resolved leaf type so that
+				// bindField uses bindKeyword directly rather than recursing
+				// through bindObject → bindDynamicObject with Type = "flattened"
+				// (which could wrap a nil emitter under race conditions).
```
shmsr:
```suggestion
				// Resolve the leaf type so bindField binds the sub-key directly
				// (e.g. keyword/long) instead of re-entering bindObject → bindDynamicObject.
```

shmsr (approved):

shmsr:
Branch is behind main after #186 (same file: `pkg/genlib/generator.go`). Please rebase onto latest main and push so this can merge. Or enable “Allow edits from maintainers” and I can update it.

